PR #176, opened 2018-04-10
Wafaa/delete k8s resources after e2e tests

### LINKED ISSUE #107: Delete all kubernetes resources after an e2e test run finishes
After e2e tests, we should clean up any kubernetes resources still around.
This only affects local dev, but it means that after running tests repeatedly you're left with a bunch of extra stuff you don't want.

WafaaT:
fixes #107 

## Reviewer Checklist

 - [x] Do you understand it?
 - [x] Is it correct?
 - [x] Is it safe?
 - [x] Is it robust?
 - [x] Is it simple?
 - [x] Is it tested?
 - [x] Is it documented?


WafaaT:
In this PR, 
- I use pytest fixtures to create setup and teardown methods for the e2e test session.
- the main objective is to clean up all the resources associated with the used namespace after the e2e   test session finishes (adding a teardown(namespace)).
- the way the tests are developed uses "namespace" as a local random name per test method, so a test_setup() is required to expose the key test parameters to the test session level.

@s1113950, currently, not all tests pass! to fix that I need your input either to fix the tests according to the new changes or use 'setup()' and teardown() per test method, which was not the main objective!

tests/conftest.py
```diff
@@ -80,3 +85,25 @@ def wrapper(path, mock=None):
         return m
 
     return wrapper
+
+@pytest.fixture
+def session_setup_teardown():
+    """ pytest setup and teardown."""
+    with create_work_dir() as workdir:
```
s1113950: We need the work dir for the whole test, so this `with` statement needs to cover everything here

tests/conftest.py
```diff
@@ -80,3 +85,25 @@ def wrapper(path, mock=None):
...
+        sess_setup = dict()
```
s1113950: This can be simplified to be:

```
app_name = str(uuid.uuid4())[:10]
sess_setup = {
    'workdir': workdir,
    'app_name': app_name,
    'namespace': getpass.getuser() + '-' + app_name
}
```
So no need to declare `dict()` as separate step

tests/conftest.py
```diff
@@ -80,3 +85,25 @@ def wrapper(path, mock=None):
...
+        sess_setup['workdir'] = workdir
+        sess_setup['app_name'] = str(uuid.uuid4())[:10]
+        sess_setup['namespace'] = getpass.getuser() + '-' \
+                                  + sess_setup['app_name']
+
+    yield sess_setup
+    
+    # remove directory when done
+    shutil.rmtree(sess_setup['workdir'])
```
s1113950: you need the old `try: finally` here; if something in the `yield` fails, then the stuff that follows won't execute:

```
In [1]: from contextlib import contextmanager

In [2]: @contextmanager
   ...: def test():
   ...:     a = 1
   ...:     yield a
   ...:     print('after the yield')
   ...:

In [3]: with test() as a:
   ...:     sdfnklfsdlksdf
   ...:
---------------------------------------------------------------------------
NameError                                 Traceback (most recent call last)
<ipython-input-3-2ae88000d893> in <module>()
      1 with test() as a:
----> 2     sdfnklfsdlksdf
      3

NameError: name 'sdfnklfsdlksdf' is not defined

In [4]:
```

Notice the `print` statement didn't happen here.

tests/conftest.py
```diff
@@ -80,3 +85,25 @@ def wrapper(path, mock=None):
...
+    # tear down remaining resources
+    current_namespace = run_popen("kubectl get jobs --namespace={}".format(
```
s1113950: `kubectl get jobs` will always return a 0 exit code :( that's what I'm fixing in another PR. You have the namespace name, I'd just try to run the namespace delete call and then if there's an error that there's no namespace handle that in an `except` block :)

tests/test_utils/e2e_commands.py
```diff
@@ -22,27 +22,27 @@
...
+        job_setup = session_setup_teardown
         # just in case tests fail, want a clean namespace always
-        self.workdir = workdir
-        self.registry = os.getenv('MLT_REGISTRY', 'localhost:5000')
-        self.registry_catalog_call = self._fetch_registry_catalog_call()
-        self.app_name = str(uuid.uuid4())[:10]
-        self.namespace = getpass.getuser() + '-' + self.app_name
-
+        self.workdir = job_setup['workdir']
+        print(" {}".format(self.workdir))
```
s1113950: leftover print?

tests/test_utils/files.py
```diff
@@ -28,6 +28,7 @@ def create_work_dir():
     try:
         yield workdir
     finally:
+    	pass
```
s1113950: This should be what it was prior

s1113950:
Also, no need to call rmtree in the setup/teardown (the create_work_dir will do this for you)

WafaaT:
@s1113950, I'm still not sure if you need this to cleanup after the whole test session or after each test! let's have a quick chat!

tests/e2e/test_deploy_flow.py
```diff
@@ -19,40 +19,72 @@
...
-    commands.undeploy()
+def test_simple_deploy(template):
+    with create_work_dir() as workdir:
+        app_params = app_parameters(workdir)
+        commands = CommandTester(app_params)
+        commands.init(template)
+        commands.build()
+        commands.deploy()
+        commands.undeploy()
+        sess_teardown(app_params['namespace'])
```
s1113950: Putting them like this won't work; if something fails then this call will never happen. It should be a fixture to be called automatically, or use sessionfinish and keep track of the namespaces as you go and delete all of them then

tests/conftest.py
```diff
@@ -18,11 +18,12 @@
 # SPDX-License-Identifier: EPL-2.0
 #
 
+from mock import MagicMock
```
dmsuehir: Was the moving of the imports intentional?  If not, please revert it.  Ebi went through and organized all the `import ...` so that they are alphabetical and then the `from ... import ...` listed after.

tests/e2e/test_deploy_flow.py
```diff
@@ -19,43 +19,72 @@
 #
 import os
 import pytest
-
+import uuid
+import getpass
```
dmsuehir: imports should be alphabetal (move `import getpass` above `import uuid`).

tests/e2e/test_deploy_flow.py
```diff
@@ -19,43 +19,72 @@
...
+from mlt.utils.process_helpers import run
```
dmsuehir: mlt imports should come after all the other `import ...` and `from ... import ...` statements (see this file as an example: https://github.com/IntelAI/mlt/blob/master/mlt/commands/build.py)

tests/e2e/test_deploy_flow.py
```diff
@@ -19,43 +19,72 @@
...
+def test_simple_deploy(template):
     with create_work_dir() as workdir:
-        commands = CommandTester(workdir)
+        app_params = app_parameters(workdir)
+        commands = CommandTester(app_params)
         commands.init(template)
         commands.build()
         commands.deploy()
         commands.undeploy()
+        sess_teardown(app_params['namespace'])
```
dmsuehir: use pytest fixtures for setup/teardown

tests/e2e/test_deploy_flow.py
```diff
@@ -19,43 +19,72 @@
...
+import uuid
+import getpass
+from mlt.utils.process_helpers import run
 from test_utils.e2e_commands import CommandTester
 from test_utils.files import create_work_dir
 
 
+# Expose the namespace value per test to be used in tearing down 
+# the remaining resources, important to have in case tests fail.
+def app_parameters(workdir):
```
dmsuehir: It seems like this app_parameters() function should be in somewhere in the `test_utils` instead of directly in this `test_deploy_flow.py` file, since it's something more general that more tests could use outside of this file in the future.  What do you think?

dmsuehir:
@WafaaT  Can the changes to `tests/conftest.py` be reverted?  It looks like things are just moved around and whitespace added?  I'm not seeing any actual changes in that file.

dmsuehir:
Hi @WafaaT 

When you sync this branch with master, can you make sure that the `commands.teardown()` is part of the `finally` that Steven added?  Also, in case you didn't see my message on slack, you'll need to resubmit this PR without the fork (just branch off master).

Thanks,
Dina

WafaaT:
Closing this, created a new one from not a forked repo to avoid circle ci test error!

PR #177, opened 2018-04-10
adding template versioning

### LINKED ISSUE #156: Add support for template versioning
Currently MLT templates do not have version. Add support for versioning MLT container templates. This will enable us to continue enhancing templates as well as smoothly updating template to new version.

For more details please see proposal: https://github.com/NervanaSystems/dls-features/blob/master/proposals/container_templates/README.md

jitendra42:
fixes #156 

## Reviewer Checklist

 - [ ] Do you understand it?
 - [ ] Is it correct?
 - [ ] Is it safe?
 - [ ] Is it robust?
 - [ ] Is it simple?
 - [ ] Is it tested?
 - [ ] Is it documented?


mlt/commands/templates.py
```diff
@@ -43,18 +43,21 @@ def action(self):
...
+                               headers=['Template', 'Version', 'Description'],
                                tablefmt="simple"))
 
     def _parse_templates(self, templates_directory):
         """parses template dirs in sorted order, pulling data from READMEs"""
         result = []
         if not os.path.exists(templates_directory):
             return result
-        for filename in sorted(os.listdir(templates_directory)):
+        for filename in sorted(next(os.walk(templates_directory))[1]):
```
s1113950: `os.walk` returns things in arbitrary order: `os.walk` under the hood calls `os.listdir` here https://docs.python.org/2/library/os.html#os.walk which returns things in arbitrary order. You'll have to ignore the file in the loop instead to be consistent (could be different on different systems)

PR #188, opened 2018-04-20
Enable versioneer for version string management before publish to PyPi

### LINKED ISSUE #187: Add a version string management for Pip installer based on the tags.
Currently and without the version string generation and management system `make dist` creates the wheel ```mlt-0.0.0-py2.py3-none-any.whl``` no matter what tag we use, meaning all versions are marked as `0.0.0`.

One way to get around this is to hardcode versions inside `setup.py` whenever we are ready to publish to `PyPi` but this way we have rely on admins to remember where and which the version is and update it prior to distribution o `PyPi`.

A cleaner way for managing versions of Python modules is using [Versioneer](https://github.com/warner/python-versioneer) and since the module is released to `Public domain`, we are free to use it.
Versioneer also prevents users from generating incorrect `PyPi` versions that later on may get rejected during `make publish`.
Versioneer is flexible enough to be directed so that it generates versions based on [PEP 440](https://www.python.org/dev/peps/pep-0440/) version formatting.

ashahba:
fixes #187 

## Reviewer Checklist

 - [x] Do you understand it?
 - [x] Is it correct?
 - [x] Is it safe?
 - [x] Is it robust?
 - [x] Is it simple?
 - [x] Is it tested?
 - [x] Is it documented?


ashahba:
@nqn and @dmsuehir Now the version for the wheel file is what it's supposed to be:
```mlt-0.1.0a1+12.gf49c412.dirty-py2.py3-none-any.whl```
Of course this is because we don't have a tag `v0.1.0-alpha2` yet. Once that is done and this one is merged, the generated wheel file will be: ```mlt-0.1.0a2-py2.py3-none-any.whl``` and that is what we publish to outside world as our first binary I guess.

File: mlt/_version.py [hunk 1]
```diff
@@ -0,0 +1,536 @@
+# Copyright (c) 2018 Intel Corporation
```
s1113950: This isn't ours; we shouldn't put an Intel corp copyright notice on it right @nqn ?

versioneer.py
```diff
@@ -0,0 +1,1822 @@
...
+
+"""The Versioneer - like a rocketeer, but for versions.
+
+The Versioneer
+==============
+
+* like a rocketeer, but for versions!
+* https://github.com/warner/python-versioneer
+* Brian Warner
+* License: Public Domain
```
s1113950: I heard that even things that are apache2 license start a 10 day review process. Is this true for `public domain` license as well @nqn ?

mlt/_version.py
s1113950: For @nqn : versioneer generates this file for us

nqn: If this is generated, can we just have it be regenerated in a release make target? That way, it won't go stale

versioneer.py
nqn: Let me ask Paritosh real quick

mlt/_version.py
s1113950: +1 to that idea

ashahba: I think we still need to put both license infos on it, since we modify it.

ashahba: Also the versioeer hasn't had a new version since early `Jan 2017` so chances are this file doesn't change very often.
Also if we want to do that overtime then we need to make it a dependency. currently we don't need that dependency and we only use the generated files which is great if you ask me.

But I'm open to ideas @nqn and @s1113950

ashahba:
@nqn let me enable testing `make dist` and then `pip install` testing for CircleCI.
Unfortunately I can't test `make release`, because it's only a one time shot 🙂

ashahba:
@nqn and @s1113950 I implemented first set of tests for `make dist` on `circleci` now.
Feel free to review and feedback otherwise we should be able to check the `is is tested` and merge this PR.

PR #193, opened 2018-04-23
Add saving app_run_id to .push.jsobn, and fix run_id bug

### LINKED ISSUE #192: Bug: There should be one run-id for entire job.
* Create single run-id for entire job, instead of creating for each deployment. 

Bug: https://github.com/IntelAI/mlt/blob/master/mlt/commands/deploy.py#L132

Move out of for loop, and at the end save in hidden file for later use.

karthikvadla:
fixes #192 

## Reviewer Checklist

 - [x] Do you understand it?
 - [x] Is it correct?
 - [x] Is it safe?
 - [x] Is it robust?
 - [x] Is it simple?
 - [x] Is it tested?
 - [x] Is it documented?

### DIFF
--- a/mlt/commands/deploy.py
+++ b/mlt/commands/deploy.py
@@ -101,12 +101,13 @@ class DeployCommand(Command):
         process_helpers.run(
             ["docker", "tag", self.container_name, self.remote_container_name])
 
-    def _update_app_run_id(self, app_run_id):
+    @staticmethod
+    def _update_app_run_id(app_run_id):
         with open('.push.json', 'r+') as json_file:
             data = json.load(json_file)
             data['app_run_id'] = app_run_id
             json_file.seek(0)
-            json.dump(data, json_file)
+            json.dump(data, json_file, indent=2)
             json_file.truncate()
 
     def _deploy_new_container(self):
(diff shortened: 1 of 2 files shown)


mlt/commands/deploy.py
```diff
@@ -122,14 +130,15 @@ def _deploy_new_container(self):
         # replaces things with $ with the vars from template.substitute
         # also patches deployment if interactive mode is set
         self.interactive_deployment_found = False
+        self.app_run_id = str(uuid.uuid4())
```
dmsuehir: why is it `self.app_run_id` versus just being `app_run_id`?  Looks like the variable is just used within this function, not anywhere else in the class.

mlt/commands/deploy.py
```diff
@@ -101,6 +101,14 @@ def _tag(self):
         process_helpers.run(
             ["docker", "tag", self.container_name, self.remote_container_name])
 
+    def _update_app_run_id(self, app_run_id):
+        with open('.push.json', 'r+') as json_file:
+            data = json.load(json_file)
+            data['app_run_id'] = app_run_id
+            json_file.seek(0)
+            json.dump(data, json_file)
```
s1113950: You're gonna want `indent=2` here so it dumps it in nice readable format :)

Also, why the `.seek(0)`? If you open as `rw` then you'll write over the file. If you open as `a`, then you'll append. Also why the .truncate()?

s1113950: Also please add a test for this function

s1113950: Ignore me for the `.seek(0)` and `.truncate()`; totally didn't get the file pointer moving and then not going back again...somehow I thought originally it would do that 😛

mlt/commands/deploy.py
karthikvadla: yeah just realized. 👍

PR #194, opened 2018-04-24
Document admin related stuff like versioneer and publish to PyPI

### LINKED ISSUE #190: Create a admin specific file under docs folder for tasks that are admin specific like PyPI release docs.
Currently for `admin` related setup like `versioneer` and `PyPI` releases there are not much documents and I've been doing most of it myself.

That need to documented so that other users can take on that responsibility if needed.
My thinking is a new document like `docs/admin_guide.md` would be a god idea.

ashahba:
fixes #190 

As I promised before to document `admin` related stuff like `versioneer` and publish to `PyPI`, here is the PR.

## Reviewer Checklist

 - [x] Do you understand it?
 - [x] Is it correct?
 - [x] Is it safe?
 - [x] Is it robust?
 - [x] Is it simple?
 - [x] Is it tested?
 - [x] Is it documented?


ashahba:
This should only go in after #188

docs/admin_guide.md
```diff
@@ -0,0 +1,128 @@
...
+repository: <repository-url>
+username: <username>
+password: <password>
+```
+
+Usually people publish the distributions to `PyPI` for latest tags.
+Once the distribution binary is built successfully for a given tag, you should be able to run:
+
+```bash
+mlt push dist/mlt-0.1.0a1+18.g60503e0-py2.py3-none-any.whl
```
dmsuehir: @ashahba is it really `mlt push`?  Not `make push`?

dmsuehir: Looks like the Makefile has a `make release` that does the twine upload.

ashahba: @dmsuehir this is typo, let me fix it.

docs/admin_guide.md
```diff
@@ -0,0 +1,128 @@
+# Administration
```
nqn: Maybe call it release.md?

docs/admin_guide.md
```diff
@@ -0,0 +1,128 @@
...
+
+## Version string management
```
nqn: If this is a reference, maybe just add to appendix? Do "release managers" have to do this again?

ashahba: No, unless under some crazy situation where the `version` files get accidentally deleted and git history is corrupted. I can't think of any other scenario.

But which `appendix` @nqn ?

nqn: But do they have to do it when they start out? if so, let's keep it here and nvm my comment

ashahba: No they really just have to do the `Creating Python distributions for PyPI` section going forward.

PR #198, opened 2018-04-25
Fix incorrect mlt --version reporting

ashahba:
After this PR the `mlt --version` should be fixed and once we have a new tag it should produce something l ike:

```
$ mlt --version
ML Container Templates Version 0.1.0-alpha1+18.gf88afdf
```

## Reviewer Checklist

 - [x] Do you understand it?
 - [x] Is it correct?
 - [x] Is it safe?
 - [x] Is it robust?
 - [x] Is it simple?
 - [x] Is it tested?
 - [x] Is it documented?


ashahba:
So when we have `v0.1.0-alpha3` tag, after this fix is merged, we should see something like:
```
$ mlt --version
ML Container Templates Version 0.1.0-alpha3
```

dmsuehir:
can we have a test for this?

ashahba:
@dmsuehir test is added next to `make dist` check in `.circleci`.

dmsuehir:
Thanks!

PR #200, opened 2018-04-26
Fix for issue #175 (specify namespace when doing kubectl exec for interactive mode)

### LINKED ISSUE #175: --interactive looks at kubectl current namespace instead of mlt.json namespace
Wafaa ran into this issue where she was getting an error about pod not found when she was using the interactive deploy.  Turns out, her kubectl was configured to use the namespace `wafaa` by default, but her `mlt.json` file had namespace set to `wstaie`.  So, it looks like `--interactive` might be assuming that the user's kubectl default namespace is the same namespace as their `mlt.json` file.  If we can instead have `--interactive` grab the namespace from the `mlt.json` file, then we can avoid user's running into this pod not found error.

```
 (.venv3) wstaie-mac01:myapp wstaie$ mlt deploy -i
file does not exits: crd-requirements.txt
Pushing  |##################################################| (ETA:  00:00:00) 
Pushed to gcr.io/constant-cubist-173123/myapp:d9f451ee-c0b2-4756-b360-45500f595e76
Deploying gcr.io/constant-cubist-173123/myapp:d9f451ee-c0b2-4756-b360-45500f595e76

$ kubectl get --namespace=wstaie all

Connecting to pod...
Retrying 1/10
Retrying 2/10
Error from server (NotFound): pods "myapp-aa86a631-d4a9-4d95-9979-1b14ddb8fcc8-jzpg2" not found
Pushing  |##################################################| (Time:  0:00:10) 
```

## Anything else we need to know?

* TBD

dmsuehir:
fixes #175

Added `--namespace` to the kubectl exec command for mlt deploy in interactive mode.  This fixes the use case where the user is using a namespace other than their default namespace.

## Reviewer Checklist

 - [ ] Do you understand it?
 - [ ] Is it correct?
 - [ ] Is it safe?
 - [ ] Is it robust?
 - [ ] Is it simple?
 - [ ] Is it tested?
 - [ ] Is it documented?

### DIFF
--- a/mlt/commands/deploy.py
+++ b/mlt/commands/deploy.py
@@ -294,5 +294,5 @@ class DeployCommand(Command):
             time.sleep(1)
 
         process_helpers.run_popen(
-            ["kubectl", "exec", "-it", podname,
+            ["kubectl", "exec", "-it", podname, "--namespace", self.namespace,
              "/bin/bash"], stdout=None, stderr=None).wait()


mlt/commands/deploy.py
```diff
@@ -294,5 +294,5 @@ def _exec_into_pod(self, podname):
             time.sleep(1)
 
         process_helpers.run_popen(
-            ["kubectl", "exec", "-it", podname,
```
nqn: Maybe add a comment of why the namespace is needed

tests/unit/commands/test_deploy.py
```diff
@@ -188,6 +188,12 @@ def test_deploy_interactive_one_file(walk_mock, progress_bar, popen_mock,
         extra_config_args={'registry': 'dockerhub'})
     verify_successful_deploy(output, interactive=True)
 
+    # verify that kubectl commands are specifying namespace
+    for call in process_helpers.run_popen.call_args_list:
+        command = call[0][0]
```
nqn: should we assert that call is a list of lists?

dmsuehir:
@nqn Regarding kubernetes_helpers.py:71 -- do you know if it's possible to install a CRD just in a namespace?  The ones that I've dealt with so far have always been cluster wide, so `kubectl get crd` returns the same thing in every namespace on the cluster.  But it wouldn't hurt to add it, just case there are namespace specific CRDs.

Issue #128, opened 2018-03-28
Add `mlt logs`  and `mlt events` command to see events and logs for the most recent run


dmsuehir:
To help users who are not familiar with `kubectl` commands, add an `mlt logs` command that displays the events and logs for pods associated with the most recent run.  By default, this command should display events/logs from all pods (color coded and/or labeled for mulit-node jobs?), but also provide the option to just see logs from a single pod.

Also, add an option to automatically start tailing logs/events after deploying (like `mlt deploy -l`).

See details in the dls-feature: https://github.com/NervanaSystems/dls-features/tree/master/proposals/abstract_infrastructure

PR #201, opened 2018-04-26
Karthik/mlt logs 128

karthikvadla:
fixes part of #128 
Adding `mlt logs` command to allow  user to view logs for the latest run.
`mlt logs` uses `kubetail` to fetch logs from multiple pods while running.

## Reviewer Checklist

 - [x] Do you understand it?
 - [x] Is it correct?
 - [x] Is it safe?
 - [x] Is it robust?
 - [x] Is it simple?
 - [x] Is it tested?
 - [x] Is it documented?


nqn:
I can start reviewing, but mind taking a look at why the CI is failing first?

mlt/commands/logs.py
```diff
@@ -0,0 +1,66 @@
...
+        super(LogsCommand, self).__init__(args)
+        self.config = config_helpers.load_config()
+
+    def action(self):
+        """
+        Display logs from all pods for latest run.
+
+        """
+
+        with open('.push.json', 'r') as f:
```
dmsuehir: What if the user has not pushed yet?  Like, if the user does `mlt init app ...`, `cd app`, `mlt logs`, then the `.push.json` file does not exist, right?  We should check for the file to exist and print a message saying something like "This application has not been deployed yet, so there are no logs to display."

mlt/commands/logs.py
```diff
@@ -0,0 +1,66 @@
...
+            data = json.load(f)
+
+        app_run_id = data['app_run_id'].split("-")
+        prefix = "-".join([self.config["name"], app_run_id[0], app_run_id[1]])
```
dmsuehir: Check for a bad `app_run_id`?  Like if `len(app_run_id) < 2`, this will break.

dmsuehir:
The issue #128  also mentions adding an option to `mlt deploy` to automatically tail the logs after deploying.  Will this be coming as part of a separate PR?

karthikvadla:
Yes Dina, i will work that as separate PR.

tests/unit/commands/test_logs.py
```diff
@@ -0,0 +1,127 @@
...
+
+    logs_command = LogsCommand({'logs': True, '--since': '1m'})
+    logs_command.config = {'name': 'app', 'namespace': 'namespace'}
+
+    with catch_stdout() as caught_output:
+        with pytest.raises(SystemExit):
+            logs_command.action()
+        output = caught_output.getvalue()
+
+    not_deployed = output.find("This app has not been deployed yet")
```
dmsuehir: This can be simplified to be like:
```
assert "This app has not been deployed yet" in output
```
(the same thing applies for the other `output.find(...)` in the other tests)

tests/unit/commands/test_logs.py
```diff
@@ -0,0 +1,127 @@
...
+
+def test_logs_kubetail_command_not_found(json_mock, open_mock, verify_init, process_helpers, os_path_mock):
+    run_id = str(uuid.uuid4())
+    os_path_mock.exists.return_value = True
+    json_mock_data = {
+        'last_remote_container': 'gcr.io/app_name:container_id',
+        'last_push_duration': 0.18889,
+        'app_run_id': run_id}
+    json_mock.load.return_value = json_mock_data
+
```
dmsuehir: too many blank lines

dmsuehir:
Can you add the logs command to the e2e tests as well?

dmsuehir:
Also, there's that `Usage Summary` section in the MLT `README.md`.  Might be good to add a call to `mlt logs` after the deploy?

README.md
```diff
@@ -109,6 +112,20 @@ Deploying gcr.io/my-project-12345/my-app:71fb176d-28a9-46c2-ab51-fe3d4a88b02c
 Inspect created objects by running:
   $ kubectl get --namespace=my-app all
 
+### Provide -l flag to tail logs immediately after deploying.
+$ mlt deploy --no-push -l                                                                                                                                                             master
+Skipping image push
+Deploying gcr.io/constant-cubist-173123/my-app:b9f124d2-ef34-4d66-b137-b8a6026bf782
```
dmsuehir: Remove our Google Cloud project name from here (see the above `mlt deploy` where I subbed in `my-project-12345` instead).

mlt/commands/deploy.py
```diff
@@ -300,3 +304,8 @@ def _exec_into_pod(self, podname):
         process_helpers.run_popen(
             ["kubectl", "exec", "-it", podname, "--namespace", self.namespace,
              "/bin/bash"], stdout=None, stderr=None).wait()
+
+    def _tail_logs(self):
+        print("Tailing logs, Please wait for few seconds...")
+        time.sleep(10)
```
dmsuehir: This doesn't seem like it will necessarily be reliable.

mlt/commands/logs.py
```diff
@@ -0,0 +1,38 @@
...
+        super(LogsCommand, self).__init__(args)
+        self.config = config_helpers.load_config()
+
+    def action(self):
+        """
+        Display logs from all pods for latest run.
+
+        """
+        print("Tailing logs, Please wait for few seconds...")
+        time.sleep(10)
```
dmsuehir: What is the sleep for?

mlt/utils/log_helpers.py
```diff
@@ -0,0 +1,71 @@
...
+import sys
+
+from mlt.utils import process_helpers
+
+def call_logs(config, args):
+    if os.path.exists('.push.json'):
+        with open('.push.json', 'r') as f:
+            data = json.load(f)
+    else:
+        print("This app has not been deployed yet,"
```
dmsuehir: add a space after the comma

mlt/utils/log_helpers.py
```diff
@@ -0,0 +1,71 @@
...
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+import json
+import os
+import sys
+
+from mlt.utils import process_helpers
+
+def call_logs(config, args):
```
dmsuehir: It would be nice if we could think of more intuitive names for these methods.  You have `call_logs` and `_get_logs` and I don't think anybody would know the difference between the two without reading the code.  At the very least, comments with descriptions of each function would help.

tests/unit/commands/test_logs.py
```diff
@@ -0,0 +1,143 @@
...
+    process_helpers.return_value.stdout.readline.return_value = ''
+    process_helpers.return_value.poll.return_value = 1
+    process_helpers.return_value.\
+        stderr.readline.side_effect = Exception(command_not_found)
+    with catch_stdout() as caught_output:
+        with pytest.raises(SystemExit):
+            logs_command.action()
+        output = caught_output.getvalue()
+
+    assert 'It is a prerequisite' in output
```
dmsuehir: There should be a blank line at the end of the file.

mlt/commands/deploy.py
karthikvadla: Oh yeah this is because, pods may taking few seconds to be online. So, if you do `kubetail` before that it just hangs. So having a sleep time will give enough time for the pods to be ready. I can reduce it to 5 secs. It will add an over head of 20 seconds to unit tests.

Any suggestions?

dmsuehir: Yeah, I think `--interactive` had the same issue.  It does a retry loop, checking if pods are up.

s1113950: Unit tests should mock sleep calls. There should be no need to wait for unit tests

README.md
```diff
@@ -109,6 +112,22 @@ Deploying gcr.io/my-project-12345/my-app:71fb176d-28a9-46c2-ab51-fe3d4a88b02c
 Inspect created objects by running:
   $ kubectl get --namespace=my-app all
 
+### Provide -l flag to tail logs immediately after deploying.
+$ mlt deploy --no-push -l                                                                                                                                                             master
+Skipping image push
+Deploying gcr.io/my-project-12345/my-app:b9f124d2-ef34-4d66-b137-b8a6026bf782
+
+Inspect created objects by running:
+$ kubectl get --namespace=kvadla all
```
dmsuehir: Change `--namespace=kvadla` to `--namespace=my-app`

README.md
```diff
@@ -121,6 +140,15 @@ $ mlt deploy -i --no-push
 Skipping image push
 Deploying localhost:5000/test:d6c9c06b-2b64-4038-a6a9-434bf90d6acc
 
+$ mlt logs                                                                                                                                                           master ✱ ◼
```
dmsuehir: what are those symbols?  The * and square?

PR #218, opened 2018-04-30
Karthik/mlt events 128

### LINKED ISSUE #128: Add `mlt logs`  and `mlt events` command to see events and logs for the most recent run

karthikvadla:
fixes #128 

## Reviewer Checklist

 - [x] Do you understand it?
 - [x] Is it correct?
 - [x] Is it safe?
 - [x] Is it robust?
 - [x] Is it simple?
 - [x] Is it tested?
 - [x] Is it documented?


mlt/commands/events.py
```diff
@@ -0,0 +1,66 @@
...
+class EventsCommand(Command):
+    def __init__(self, args):
+        super(EventsCommand, self).__init__(args)
+        self.config = config_helpers.load_config()
+
+    def action(self):
+        """
+        Display events for the latest run
+        """
+        with open('.push.json', 'r') as f:
```
dmsuehir: If the user hasn't deployed yet, then this file does not exist.  We need to check for that and display a proper message, like "This application has not been deployed yet, so there are no events to display."

mlt/commands/events.py
```diff
@@ -0,0 +1,66 @@
...
+            data = json.load(f)
+
+        app_run_id = data['app_run_id'].split('-')
+        filter_tag = "-".join([self.config["name"],
+                               app_run_id[0],
```
dmsuehir: is there a reason why you are grabbing 4 chunks from the app run id, where as in the logs you grab 2?  Do the events show the job name (which has the full app_run_id) or the pod name (which has a truncated app_run_id, depending on how long the app name is)?

mlt/commands/events.py
```diff
@@ -0,0 +1,66 @@
...
+                               app_run_id[1],
+                               app_run_id[2],
+                               app_run_id[3]])
+        namespace = self.config['namespace']
+        self.get_events(filter_tag, namespace)
+
+    @staticmethod
+    def get_events(filter_tag, namespace):
+
+        awk = 'awk \'NR=1 || /{}/\''.format(filter_tag)
```
dmsuehir: If there are no events, what does the output look like?

mlt/commands/events.py
```diff
@@ -0,0 +1,66 @@
...
+        events_cmd = "kubectl get events " \
+                     "--namespace {} | {}"\
+            .format(namespace, awk)
+
+        events = process_helpers.run_popen(events_cmd, shell=True)
```
dmsuehir: Since we are just getting and displaying the output, it seems like it would be simpler to use `subprocess.check_output` (then you don't need the while loop)

mlt/commands/events.py
```diff
@@ -0,0 +1,66 @@
...
+
+        while True:
+            output = events.stdout.read(1)
+            if output == '' and events.poll() is not None:
+                break
+            if output is not '':
+                sys.stdout.write(output)
+                sys.stdout.flush()
```
dmsuehir: missing blank line at the end of the file

mlt/commands/events.py
karthikvadla: Yes for events while parsing, it expects complete name. 
whereas for logs..just a prefix is good.

mlt/commands/events.py
```diff
@@ -0,0 +1,94 @@
...
+        """
+         Fetches events
+        """
+        events_cmd = ["kubectl", "get", "events", "--namespace", namespace]
+        try:
+            events = process_helpers.run_popen(events_cmd)
+            first_line = True
+            header = events.stdout.readline()
+            while True:
+                output = events.stdout.readline()
```
dmsuehir: If there are no events, what does `mlt events` show?

tests/unit/commands/test_events.py
```diff
@@ -0,0 +1,125 @@
...
+    json_mock_data = {
+        'last_remote_container': 'gcr.io/app_name:container_id',
+        'last_push_duration': 0.18889,
+        'app_run_id': run_id}
+    json_mock.load.return_value = json_mock_data
+
+    events_command = EventsCommand({'events': True})
+    events_command.config = {'name': 'app', 'namespace': 'namespace'}
+
+    process_helpers.side_effect = OSError
```
dmsuehir: I would add a message to the error `OSError("expected error message")` and then at the end verify `assert "expected error message" in output` to ensure that we are displaying the error.

tests/unit/commands/test_events.py
```diff
@@ -0,0 +1,125 @@
...
+
+@pytest.fixture
+def process_helpers(patch):
+    return patch('process_helpers.run_popen')
+
+@pytest.fixture
+def verify_init(patch):
+    return patch('config_helpers.load_config')
+
+def test_events_get_events(json_mock, open_mock, verify_init, process_helpers, os_path_mock):
```
dmsuehir: Also add a test for when there are no events (where the the output of `kubectl get events` is just a single line saying "No resources found.") to ensure we are properly handling that case too.

mlt/commands/events.py
```diff
@@ -0,0 +1,94 @@
...
+        self._get_events(filter_tag, namespace)
+
+    @staticmethod
+    def _get_events(filter_tag, namespace):
+        """
+         Fetches events
+        """
+        events_cmd = ["kubectl", "get", "events", "--namespace", namespace]
+        try:
+            events = process_helpers.run_popen(events_cmd)
```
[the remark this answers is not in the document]
karthikvadla: Nope, as i'm doing `kubectl get events -n kvadla`. and then i need to parse each line to filter with run_id. run_popen gives me stream, it's easy to read each line and check.

README.md
```diff
@@ -127,6 +127,10 @@ $ kubectl get --namespace=robertso all
 Connecting to pod...
 root@test-9e035719-1d8b-4e0c-adcb-f706429ffeac-wl42v:/src/app# ls
 Dockerfile  Makefile  README.md  k8s  k8s-templates  main.py  mlt.json	requirements.txt
+
+# Displays events for the current job
+$ mlt events
+
```
dmsuehir: Can you give some output here as an example of what the user might see after doing the deploy?

Issue #75, opened 2018-03-13
Add support to update existing mlt project with updated template


nqn:
It would be great if we could benefit from continued development on templates.
Say that enhancements are made to the dockerfiles, convenience and debugging scripts are added to the templates. Having to start over with `mlt init` is not a great experience, so suggesting that we provide a:

```
$ mlt init -t=tf-distributed my-app
$ cd my-app
# ... changes happen in the `tf-distributed` template
$ mlt update
# ... Apply changes; if any conflicts, keep the original one and have a .orig, .new and .digg
```

PR #225, opened 2018-05-04
Add mlt status command

### LINKED ISSUE #127: Add `mlt status` command for monitoring job status
As part of the dls-feature that helps users who are not familiar with `kubectl` commands, an `mlt status` command that shows the basic status of pods associated with the last run (containers creating, running, error, etc).

See details in the dls-feature: https://github.com/NervanaSystems/dls-features/tree/master/proposals/abstract_infrastructure

dmsuehir:
fixes #127

Adds `mlt status` command.  

Finding the pods associated with a job can be different based on which template is used.  For example, with hello-world we can look for pods with a label `job-name=...`.  But for `TFJob` templates, we instead need to look for pods with a label `tf_job_name=...`.  Also, since the TFJob pods are terminated after the job completes, we should also show list the `TFJob` in the status (otherwise the status doesn't show anything after the job completes).  Since these commands are different per template, I added a `status` rule to the `Makefile` for each template.  And in case the user has a template that doesn't have the `status` rule in the `Makefile` (like maybe they have an old app from before this was added), then I have a default way to grab status based on pod name.

## Reviewer Checklist

 - [ ] Do you understand it?
 - [ ] Is it correct?
 - [ ] Is it safe?
 - [ ] Is it robust?
 - [ ] Is it simple?
 - [ ] Is it tested?
 - [ ] Is it documented?


mlt/commands/status.py
```diff
@@ -0,0 +1,94 @@
...
+            raise ValueError("Unable to get job status because the run id {} "
+                             "is invalid".format(app_run_id))
+
+        # Pod name prefix for filtering (we can't filter by pod label, since
+        # the label name is different depending on job type)
+        pod_name_prefix = "-".join([app_name, run_id_list[0], run_id_list[1]])
+
+        # Get pods list and display it
+        cmd = "kubectl get pods -a -o wide --namespace {} | " \
+              "awk 'NR==1 || /{}/'".format(namespace, pod_name_prefix)
```
nqn: Let's try to the parsing in python and not pipe to `awk`, _and_ get rid of `shell=True`. In general, assume that any subprocess call is as sensitive as sql query strings. Someone could set the namespace in the config file to `&& rm -rf /` or something worse.

mlt/commands/status.py
```diff
@@ -0,0 +1,94 @@
...
+                # Job has probably been undeployed.  Remove make error and
+                # display the output.
+                make_error = "make: *** [status] Error 1"
+                str_output = str(e.output).replace(make_error, "").strip()
+                print(str_output)
+            else:
+                print("Error while getting app status: {}".format(e.output))
+
+    @staticmethod
+    def _default_status(namespace, app_name, app_run_id):
```
nqn: How about we just don't do anything if there isn't a target for it?

dmsuehir: @nqn  My thought was that if someone had initialized an app prior to `mlt status` existing, then their app won't have the `status` target in the Makefile, so I figured some default behavior to get the pods for their job would at least get them some status information.  But, if we are ok with not supporting that use case, I am fine with removing this and instead printing a message about the command being unsupported for their app.  What we really need is #75 so that they can grab and update.  :)

dmsuehir:
@nqn I made some updates:
* Removed the "default" status behavior (when the target isn't found in the Makefile) and instead just print a message saying that the app does doesn't support the status command.
* Removed the check for the specific makefile error and instead handle it from the Makefile itself.
* Remove the `shell=True`

PR #233, opened 2018-05-11
New tf-distributed which uses MonitoredTrainingSession

tonyreina:
wip

The Distributed TensorFlow projects now recommend using a MonitoredTraining session. This automatically handles the checkpoints, failover, and coordination between the nodes. 

I've re-written the tf-distributed template to use a Monitored Session. I've also broken out the model code into a file called model.py. Right now it is just done as a dictionary; Krishna has a class implementation that it better (I'm going to eventually migrate to it, but wanted to push the code for review now since the monitored session is important.)

Only known bug is with the checkpoints directory. I think I need to add KVC to allow the code to use a common checkpoint directory for all nodes. I'm not sure how to do this.

## Reviewer Checklist

 - [ ] Do you understand it?
 - [ ] Is it correct?
 - [ ] Is it safe?
 - [ ] Is it robust?
 - [ ] Is it simple?
 - [ ] Is it tested?
 - [ ] Is it documented?

### DIFF
--- a/mlt-templates/tf-dist-mnist/data.py
+++ b/mlt-templates/tf-dist-mnist/data.py
@@ -1,0 +1,60 @@
+#
+# -*- coding: utf-8 -*-
+#
+# Copyright (c) 2018 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+
+
+import tensorflow as tf
+import numpy as np
+import logging
+
+
+def load_data():
+
+    # Load pre-shuffled MNIST data into train and test sets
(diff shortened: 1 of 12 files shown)


tonyreina:
I've also re-written tf-dist-mnist in the same fashion. In this case I refactored the data loader as a separate script (data.py).  Again, I like Krishna's way to handling this as a python class rather than a dictionary (which is what I did). However, this code works and is fairly simple to follow. Hopefully, someone can help convert the dictionary to a class??

Only error I still get is with the checkpoints. I think I need to setup KVC and declare the checkpoint_directory as a NFS or S3 folder. I'm hoping someone can help with that part too.

Current code works but doesn't save checkpoints. Also, I'll hopefully get to add TensorBoard images in the next version. If the checkpoint directory is setup properly, then regular TensorBoard stuff should work.

tonyreina:
Just wanted to see if someone could please review. This will definitely make the distributed templates easier to use for the data scientists since it refactors the main.py into a model.py and data.py.

mas-dse-greina:
Just bumping this again to see if someone can review for the PR. Thanks.

-Tony

ashahba:
@tonyreina Thanks for the PR.
Ping me directly so we can fix the PR and have it pass the tests before we can review and merge it.

Thanks.

s1113950:
@tonyreina @ashahba what's the status of this?

ashahba:
@s1113950 I'll review and update here.

tonyreina:
Thanks. Also, when you update it can you remove the `import logging` and change the `logging.info` statements to `tf.logging.info` statements. That should prevent duplicate logs.

s1113950:
@ashahba what's the status of this?

nqn:
Closing due to project deprecation

PR #234, opened 2018-05-14
Add mlt config command and tests

### LINKED ISSUE #129: Add `mlt config` command to view/modify parameters in the mlt.json file
Add `mlt config` command to make it easier for new users to view/change parameters in the mlt.json file.  `mlt config list` can display all the current parameters/values.  Then to update parameters, `mlt config set [parameter name] [new value]` can be used.

See details in dls-feature:
https://github.com/NervanaSystems/dls-features/tree/master/proposals/abstract_infrastructure

dmsuehir:
fixes #129

Add `mlt config` command (with subcommands: list, set, remove) to allow users to view and modify values in the `mlt.json` file.

## Reviewer Checklist

 - [ ] Do you understand it?
 - [ ] Is it correct?
 - [ ] Is it safe?
 - [ ] Is it robust?
 - [ ] Is it simple?
 - [ ] Is it tested?
 - [ ] Is it documented?

### DIFF
--- a/mlt/commands/config.py
+++ b/mlt/commands/config.py
@@ -1,0 +1,123 @@
+#
+# -*- coding: utf-8 -*-
+#
+# Copyright (c) 2018 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software`
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+
+import sys
+from tabulate import tabulate
+
+from mlt.commands import Command
+from mlt.utils import config_helpers, constants
+
+
+class ConfigCommand(Command):
+    def __init__(self, args):
+        super(ConfigCommand, self).__init__(args)
+        self.list = args.get('list', False)
+        self.set = args.get('set', False)
+        self.unset = args.get('unset', False)
+        self.name = args.get('<name>', None)
+        self.value = args.get('<value>', None)
+        self.config = config_helpers.load_config()
+
+        self.param_keys = self.name.split('.') if self.name else None
+
+        if (self.set or self.unset) and not self.name:
+            print("Name of the configuration parameter must be specified.")
+            sys.exit(1)
+
+    def action(self):
+        # Call the specified sub-command
+        if self.list:
+            print("list")
+            self._list_configs()
+        elif self.set:
+            print("set")
+            self._set_config()
+        elif self.unset:
+            print("unset")
+            self._unset_config()
+        else:
+            print("Invalid config command.  Must have one of the following"
+                  "sub-command selected: list, set, unset.")
+            sys.exit(1)
+
+    def _list_configs(self):
+        """
+        Lists the config parameters and their values using tabulate.
+        """
+        config_list = []
+        for config_name in self.config.keys():
+            if config_name == constants.TEMPLATE_PARAMETERS:
+                template_params = self.config[config_name]
+                for template_config in template_params.keys():
+                    config_list.append(["{}.{}".format(
+                        constants.TEMPLATE_PARAMETERS, template_config),
+                        template_params[template_config]])
+            else:
+                config_list.append([config_name, self.config[config_name]])
+
+        if len(config_list) > 0:
+            print(tabulate(config_list, headers=['App Config', 'Value']))
+        else:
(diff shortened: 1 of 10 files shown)


mlt/commands/config.py
```diff
@@ -0,0 +1,123 @@
...
+            print("list")
+            self._list_configs()
+        elif self.set:
+            print("set")
+            self._set_config()
+        elif self.unset:
+            print("unset")
+            self._unset_config()
+        else:
+            print("Invalid config command.  Must have one of the following"
```
s1113950: This should never happen; `docopt` will error. Ex:
```
robertso@robertso-mac02:~/test$ mlt config pasta
Usage:
  mlt (-h | --help)
  mlt --version
  mlt init [--template=<template> --template-repo=<repo>]
      [--registry=<registry> --namespace=<namespace]
      [--skip-crd-check] <name>
  mlt config (list | set <name> <value> | unset <name>)
  mlt build [--watch]
  mlt deploy [--no-push] [-i | --interactive]
      [--retries=<retries>] [--skip-crd-check] [<kube_spec>]
  mlt undeploy
  mlt status
  mlt (template | templates) list [--template-repo=<repo>]
```
so no need for this `else`

mlt/commands/config.py
```diff
@@ -0,0 +1,123 @@
...
+            if config_name == constants.TEMPLATE_PARAMETERS:
+                template_params = self.config[config_name]
+                for template_config in template_params.keys():
+                    config_list.append(["{}.{}".format(
+                        constants.TEMPLATE_PARAMETERS, template_config),
+                        template_params[template_config]])
+            else:
+                config_list.append([config_name, self.config[config_name]])
+
+        if len(config_list) > 0:
```
s1113950: `if not config_list` is the recommended way to check if empty list:

`For sequences, (strings, lists, tuples), use the fact that empty sequences are false.` from here:
https://www.python.org/dev/peps/pep-0008/

mlt/commands/config.py
```diff
@@ -0,0 +1,123 @@
...
+            print(tabulate(config_list, headers=['App Config', 'Value']))
+        else:
+            print("No configuration parameters to display.")
+
+    def _set_config(self):
```
s1113950: Can we pass in `param_keys` to this function? Not immediately clear what `the config parameter` is exactly.

mlt/commands/config.py
```diff
@@ -0,0 +1,123 @@
...
+        """
+        Sets the config parameter to the specified value.  If the config
+        parameter does not already exist in the MLT config, then it's added.
+        Writes the new config back to the mlt config file.
```
s1113950: Since we're always adding it, why not do something like:

```
self.config['config_parameter'] = value
config_helpers.update_config(self.config)
```
?
From the `mlt config set` command it looks like we only add 1 at a time right?

mlt/commands/config.py
```diff
@@ -0,0 +1,123 @@
...
+        matched_config[self.param_keys[-1]] = self.value
+        config_helpers.update_config(self.config)
+
+    def _unset_config(self):
+        """
+        Finds, then removes the specified config parameter from the MLT config
+        file.  If the config parameter does not exist in the MLT config file,
+        then an error is displayed.
+        """
+        matched_config = self.config
```
s1113950: Same with here, couldn't it be a try except around a `del self.config['config_parameter']` call to handle the key not found error, followed by the `update_config` call? Perhaps I'm missing something

mlt/main.py
```diff
@@ -97,7 +100,7 @@ def sanitize_input(args, regex=None):
        It is recommended on docopt github to do validation
     """
     # docker requires repo name to be in lowercase
-    if args["<name>"]:
+    if args["<name>"] and args.get("init", None):
```
s1113950: by default `.get()` returns `None` if not found, so no need for the second param here

mlt/commands/config.py
```diff
@@ -0,0 +1,123 @@
...
+        self.param_keys = self.name.split('.') if self.name else None
+
+        if (self.set or self.unset) and not self.name:
+            print("Name of the configuration parameter must be specified.")
+            sys.exit(1)
+
+    def action(self):
+        # Call the specified sub-command
+        if self.list:
+            print("list")
```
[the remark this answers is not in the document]
dmsuehir: 😂 Totally did not mean to leave those in.

mlt/commands/config.py
dmsuehir: We are only setting one at time, but the configs are in json (`mlt.json`), so they could be nested, and so we can't necessarily just set it like `self.config['config_parameter'] = value` (it might be `self.config['thing1']['thing2']['config_parameter'] = value`).

mlt/commands/config.py
dmsuehir: Same issue here, it's because of the possibility of nested configs.

mlt/commands/config.py
s1113950: ahh ok, and that's what `param_keys` could look like. Ok then! Is there an easy way to combine logic for checking for key existence then? It looks nearly identical to logic in `_unset_config`

PR #249, opened 2018-05-23
Fix working with local template repos (issue #154)

### LINKED ISSUE #154: Specifying a local `--template-repo` does not always work
We should be able to use the `--template-repo` arg to specify a local path with `mlt init` and `mlt template list`, but this does not always work.  For example, if I create a new template directory, it's not showing up in `mlt template list` unless I commit the new files to git.

For example, here I have added a new template directory `mlt-templates/taco-world` which is just a copy of `hello-world` and it's not showing up when I specify the local mlt directory for `--template-repo`.

```
dmsuehir-mac01:mlt dmsuehir$ git status
On branch dina/new_template
Untracked files:
  (use "git add <file>..." to include in what will be committed)

	.DS_Store
	.idea/
	mlt-templates/.DS_Store
	mlt-templates/taco-world/

nothing added to commit but untracked files present (use "git add" to track)
dmsuehir-mac01:mlt dmsuehir$ mlt templates list --template-repo=$(PWD)
Template        Description
--------------  --------------------------------------------------------------------------------------------------
hello-world     A TensorFlow python HelloWorld example run through Kubernetes Jobs.
pytorch         Sample distributed application taken from http://pytorch.org/tutorials/intermediate/dist_tuto.html
tf-dist-mnist   A distributed TensorFlow MNIST model which designates worker 0 as the chief.
tf-distributed  A distributed TensorFlow matrix multiplication run through the TensorFlow Kubernetes Operator.
```

dmsuehir:
fixes #154

When the `--template-repo` is not a git repo, copy the local repo over the git clone so that we don't lose any uncommitted files.  This is done in `git_helpers.clone_repo` and is used for both `mlt template list` and `mlt init`.

## Reviewer Checklist

 - [ ] Do you understand it?
 - [ ] Is it correct?
 - [ ] Is it safe?
 - [ ] Is it robust?
 - [ ] Is it simple?
 - [ ] Is it tested?
 - [ ] Is it documented?

### DIFF
--- a/mlt/utils/git_helpers.py
+++ b/mlt/utils/git_helpers.py
@@ -22,6 +22,7 @@ import os
 import shutil
 import tempfile
 from contextlib import contextmanager
+from distutils.dir_util import copy_tree
 
 from mlt.utils import process_helpers
 
@@ -32,6 +33,12 @@ def clone_repo(repo):
     process_helpers.run_popen(
         "git clone {} {}".format(repo, destination),
         shell=True, stdout=False, stderr=False).wait()
+
+    # If the template repo is a local path, then copy the local directory over
+    # the git clone so that the templates reflect the local changes.
+    if not is_git_repo(repo):
+        copy_tree(repo, destination)
+
     try:
         yield destination
     finally:
@@ -39,3 +46,9 @@ def clone_repo(repo):
         # https://bugs.python.org/issue29699
         if os.path.exists(destination):
             shutil.rmtree(destination)
+
+
+def is_git_repo(template_repo):
+    """ Returns True if the template_repo looks like a git repository. """
+    return template_repo.startswith("git@") or \
+        template_repo.startswith("https://")
--- a/tests/e2e/test_templates.py
+++ b/tests/e2e/test_templates.py
@@ -27,10 +27,14 @@ from test_utils.e2e_commands import CommandTester
 from test_utils.files import create_work_dir
 
 
+def call_template_list(template_repo):
+    return check_output(['mlt', 'templates', 'list',
+                         '--template-repo={}'.format(template_repo)]
+                        ).decode("utf-8")
+
+
 def test_templates():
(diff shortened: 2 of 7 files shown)


mlt/utils/git_helpers.py
```diff
@@ -22,6 +22,7 @@
 import shutil
 import tempfile
 from contextlib import contextmanager
+from distutils.dir_util import copy_tree
```
s1113950: how does this differ from `shutil.copytree`? We use that in `init.py`. It could be good to unify them if possible?

tests/e2e/test_templates.py
```diff
@@ -35,3 +40,46 @@ def test_templates():
...
+        new_template_name = "test-local"
+        new_template_directory = os.path.join(templates_directory,
+                                              new_template_name)
+        os.mkdir(new_template_directory)
+        new_template_file = os.path.join(new_template_directory, "README.md")
+        with open(new_template_file, "w") as f:
+            f.write("New local template for testing")
+
+        # call mlt template list and then check the output
+        output = check_output(['mlt', 'templates', 'list',
```
s1113950: This is the same call as the other `test_templates` `mlt templates list` function; could you helper function this for when we add new `mlt template list` tests in the future?

mlt/utils/git_helpers.py
dmsuehir: The difference is that copytree doesn't overwrite files.  I started out by trying to use that, but had to move to `copy_tree`.

tests/e2e/test_templates.py
dmsuehir: :yeah:

dmsuehir: @s1113950 I moved the `mlt templates list` call to a function.

dmsuehir:
closing this PR because I opened a PR without the fork (#251)

PR #257, opened 2018-05-25
Use .gitignore file instead of empty README.md in empty directories.

ashahba:
…ME.md files

Currently we use blank `README.md` files within `k8s` directory for each template mainly to get around adding empty folders to `git`.
But an approved solution is provided here:
https://stackoverflow.com/questions/115983/how-can-i-add-an-empty-directory-to-a-git-repository

So the idea with this PR is to do it the way others do 🙂 and they can later understand better what this directory is and the contents may come later.

## Reviewer Checklist

 - [ ] Do you understand it?
 - [ ] Is it correct?
 - [ ] Is it safe?
 - [ ] Is it robust?
 - [ ] Is it simple?
 - [ ] Is it tested?
 - [ ] Is it documented?


karthikvadla:
LGTM :)

PR #261, opened 2018-05-26
Add a new template for tensorboard.

WafaaT:
fixes #228 

## Reviewer Checklist

 - [ ] Do you understand it?
 - [ ] Is it correct?
 - [ ] Is it safe?
 - [ ] Is it legally compliant?
 - [ ] Is it robust?
 - [ ] Is it simple?
 - [ ] Is it tested?
 - [ ] Is it documented?
 - [ ] Is the style consistent?

See [the reviewer guide](docs/reviews.md) for more detail on each check.

PR #264, opened 2018-05-29
Add a new template for tensorboard.

WafaaT:
fixes #261 

## Reviewer Checklist

 - [x] Do you understand it?
 - [x] Is it correct?
 - [x] Is it safe?
 - [x] Is it robust?
 - [x] Is it simple?
 - [x] Is it tested?
 - [x] Is it documented?


mlt-templates/tensorboard/Makefile
```diff
@@ -0,0 +1,33 @@
...
+
+main: main.py
+	@echo "Should run linting and tests before submitting application"
+
+build:
+	docker build --build-arg HTTP_PROXY=${HTTP_PROXY} --build-arg HTTPS_PROXY=${HTTPS_PROXY} --build-arg http_proxy=${http_proxy} --build-arg https_proxy=${https_proxy} -t ${CONTAINER_NAME} .
+
+status:
+	@echo "TB Job:"
+	-@kubectl get tbjob ${JOB_NAME} --namespace ${NAMESPACE} || echo "The job may have been undeployed."
```
dmsuehir: Does this work?  It looks like this is just using a kubernetes deployment, so I would have thought that getting the status would look more like `kubectl get deployment ...`?

mlt-templates/tensorboard/k8s-templates/tbjob.yaml
```diff
@@ -0,0 +1,29 @@
...
+  template:
+    metadata:
+      labels:
+        mlt-app-name: $app
+    spec:
+      containers:
+        - image: $image
+          name: tensorflow
+          command : ["/bin/sh","-c"]
+          args : ["tensorboard --logdir $path; run -d debian sleep 300"]
```
dmsuehir: what happens after this sleep expires?  does the tensorboard session die?

mlt-templates/tensorboard/parameters.json
```diff
@@ -0,0 +1,5 @@
+{
+  "template_parameters" : [
+    { "name": "path", "value": "path to tensorboard logs directory" }
```
dmsuehir: Maybe for the `value` here we could give the user a better hint as to what goes here.  For example, what if we have the value as `s3://your-bucket/log-directory`.

mlt-templates/tensorboard/Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+
+FROM python:3
+
+ADD requirements.txt /src/deps/requirements.txt
+RUN pip install -r /src/deps/requirements.txt
+
+WORKDIR /src/app
+ADD . /src/app
+
+RUN pycodestyle -v .
```
dmsuehir: This isn't needed, we don't have any python to validate, right?

mlt-templates/tensorboard/Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+
+ENTRYPOINT [ ]
```
dmsuehir: I don't think this line is necessary?

mlt-templates/tensorboard/requirements.txt
```diff
@@ -0,0 +1,3 @@
+tensorflow==1.4.0
+pycodestyle==2.3.1
```
dmsuehir: we don't need this since there's no python to validate.

mlt-templates/tensorboard/requirements.txt
```diff
@@ -0,0 +1,3 @@
...
+scipy==1.0.1
```
dmsuehir: is `scipi` needed for TensorBoard?

dmsuehir:
This is a little different from the other templates in that it doesn't work out-of-the-box.  It requires that the user updates the `path` to point to their log directory.  I'm wondering if this is okay, or if we should have some sample checkpoint files (like from an MNIST model training or something) that we can point to as the default.

Also, I'm wondering what happens if tensorboard is given a bad path.  Does it show as an error in kubernetes?  Because we have a generic test that tries to deploy all the templates.

mlt-templates/tensorboard/Makefile
WafaaT: Yes, it's not a tbjob.. I would use a deployment or service instead! but in this case what would the variable `${JOB_NAME}` return. I mean based on my yaml file, how can I edit this value or define a new one!

mlt-templates/tensorboard/Makefile
```diff
@@ -0,0 +1,33 @@
...
+	@echo ""
+	@echo "Pods:"
+	@kubectl get pods --namespace ${NAMESPACE} -o wide -a -l tb_job_name=${JOB_NAME}
```
WafaaT: also for this command, the deployed application here deploys a pod with the same name of the deployment but I think with random UUID at the end.
for example: 
`deploy/tb-big-tb-test-e88b2177-4218-470d-99c6-fdec83acd737` and the pod: `po/tb-big-tb-test-e88b2177-4218-470d-99c6-fdec83acd737-5b86dd6zr8k`. then how to handle this case here?

mlt-templates/tensorboard/k8s-templates/tbjob.yaml
WafaaT: this is only to keep the pod there for a while (it's not a clean way though). In my tests, the service, deployment and pod are there until I undeploy or delete it manually. 
I talked to @dzungductran and @ashahba . and then agreed on using it or while loop and refer to the open issue in kubeflow https://github.com/kubeflow/tf-operator/issues/536

dmsuehir: but we aren't using the tf-operator here, so I don't understand how that other issue applies?

mlt-templates/tensorboard/parameters.json
WafaaT: I think if we used specific format like `s3://your-bucket/log-directory` would tell users that `s3` storage is what we support! I thought we would leave it up to the user storage method and how they access it. what do you think?

mlt-templates/tensorboard/Makefile
dmsuehir: ${JOB_NAME} is the app-runid (see https://github.com/IntelAI/mlt/blob/master/mlt/commands/status.py#L47).  It looks like you are naming the deployment "tb-$app-$run" so I think you should be able to do something like `kubectl get deployment tb-${JOB_NAME} --namespace ${NAMESPACE}`

mlt-templates/tensorboard/Makefile
dmsuehir: The `${JOB_NAME}` environment variable includes the uuid.  You just need to know how the pod is labeled (currently this line is looking for a label named `tf_job_name`... is this correct?)

mlt-templates/tensorboard/k8s-templates/tbjob.yaml
WafaaT: I would just memic what they would come up with here just to keep the pod a life as needed. so this is a temp. solution for now

mlt-templates/tensorboard/parameters.json
dmsuehir: I would suggest calling this `logdir` instead of path, since that's the name of the tensorboard parameter.

mlt-templates/tensorboard/Makefile
WafaaT: Let's chat about the part as I'm a bit confused :)

mlt-templates/tensorboard/Makefile
WafaaT: how a bout the service case? do I need to add a command for it as well? or just add one for the service as it's our target?

WafaaT:
@dmsuehir I tested it against bad path values, and without sample checkpoints and it does not error out in kubernetes. what happens is that when we open TB in the browser it tells that there's nothing to display for the given path!

mlt-templates/tensorboard/README.md
```diff
@@ -0,0 +1,82 @@
+# TensorBoard MLT Application
+
+A Tensorboard service in kubernetes cluster.
+
+This template requires the path to the TensorFlow job logs to be configured.
+The command below shows how to update the log directory value in `k8s-templates/tbjob.yaml`.
+
+`mlt config set template_parameters.logdir <value>`.
```
dmsuehir: I would remove the . at the end of this line, because someone might accidentally copy/paste that and think it's part of the command.

mlt-templates/tensorboard/README.md
```diff
@@ -0,0 +1,82 @@
+# TensorBoard MLT Application
+
+A Tensorboard service in kubernetes cluster.
```
dmsuehir: Capitalize the "B" in TensorBoard, for consistency, and it looks like that's how they have it in GitHub.  Also, capitalize the "K" in Kubernetes.

dmsuehir: I see other places in this doc with the same capitalization issues, so probably do a find and and replace this whole file.

mlt-templates/tensorboard/README.md
```diff
@@ -0,0 +1,82 @@
...
+
+This template requires the path to the TensorFlow job logs to be configured.
+The command below shows how to update the log directory value in `k8s-templates/tbjob.yaml`.
+
+`mlt config set template_parameters.logdir <value>`.
+
+Tensorboard template works with different storage sources, and it's the user responsibility to allow the TensorBoard service to access the logs location.
+
+This template assumes that Ingress controller is enabled in the kubernetes cluster.
```
dmsuehir: Since Ingress is required, I think we should add it to the list of pre-requisites for the TensorBoard template (https://github.com/IntelAI/mlt#build)

mlt-templates/tensorboard/README.md
```diff
@@ -0,0 +1,82 @@
...
+
+Tensorboard template works with different storage sources, and it's the user responsibility to allow the TensorBoard service to access the logs location.
+
+This template assumes that Ingress controller is enabled in the kubernetes cluster. 
+
+
+
+## Example: Using the TensorBoard template with Google Cloud Storage
+
+Assuming that the user stores TensorFLow jobs logs/results in gcloud storage as S3 bucket, then the 'logdir' value would be
```
dmsuehir: TensorFLow ➡️ TensorFlow

mlt-templates/tensorboard/README.md
```diff
@@ -0,0 +1,82 @@
...
+## Example: Using the TensorBoard template with Google Cloud Storage
+
+Assuming that the user stores TensorFLow jobs logs/results in gcloud storage as S3 bucket, then the 'logdir' value would be 
+`S3://<user-bucket-name>/<directory-name>`.
+
+In this case, to allow access to the storage source, the user gets the access key and ID from gcloud storage and use them to create
+`gcs-creds` secrets in `kubectl`:
+
+```
+GCR_ACCESS_KEY_ID=user-access-key-id-value
```
dmsuehir: This is probably my fault cause I think I gave you this snippet before, but can you change the variables to be named GCS instead of GCR (since it's for Google Cloud _Storage_, not Google Cloud Registry).

mlt-templates/tensorboard/README.md
```diff
@@ -0,0 +1,82 @@
...
+ ```
+
+### Steps to launch TensorBoard service using TensorBoard MLT template:
+
+1. Create a TensorBoard application using the TensorBoard MLT template.
+`mlt init tensorboard-app --template=tensorboard`
+
+From inside the application `cd tensorboard-app` :
+
+2. Set TensorbBoard `logdir`.
```
dmsuehir: TensorbBoard ➡️ TensorBoard

mlt-templates/tensorboard/README.md
```diff
@@ -0,0 +1,82 @@
...
+`mlt config set template_parameters.logdir S3://<user-bucket-name>/<directory-name>`
+
+3. Create the application image.
+`mlt build`
+
+In this example, the user need to modify k8s-templates/tbjob.yaml to add the previously created secrets, as follows:
+
+```
+env:
```
dmsuehir: Add the proper indentation in front of `env:`

mlt-templates/tensorboard/README.md
```diff
@@ -0,0 +1,82 @@
...
+          - name: S3_VERIFY_SSL
+            value: "true"
+          - name: S3_ENDPOINT
+            value: "storage.googleapis.com"
+```
+
+4. Deploy TensorBoard service.
+`mlt deploy` 
+
+5. Use the given command `kubectl get --namespace=<user-namespace> all`
```
dmsuehir: hmm... can `mlt status` be used instead?  If so, we should say to use that instead of the `kubectl` command.

mlt-templates/tensorboard/README.md
dmsuehir: Oh, and I just read your note at the bottom about GKE not needing this.  Maybe we should just mention here that Ingress is not needed for GKE, otherwise someone might not find that out until the end.

PR #266, opened 2018-05-29
Threaded tests

s1113950:
- Changes tests from function to class-based. Easier to thread and manage logic
- bumps resource class to `xlarge`, I have a call with circleci later today to allow this for mlt too (we have it for nprof)
- removes `gcloud docker` call because now docker updated itself and tests won't pass anymore because `gcloud docker` isn't supported past docker version 18.03.
- We no longer have `create_workdir()` contextmanger weirdness, and manage this from the pytest conftest instead
- lots of tweaks to e2e_commands to allow for threading taking extra time in some places (some tests assumed that images were already built, like --watch flag, as that always ran before after the test that built the image for it, so we now have to wait more for that to potentially build stuff)

## Reviewer Checklist

 - [ ] Do you understand it?
 - [ ] Is it correct?
 - [ ] Is it safe?
 - [ ] Is it robust?
 - [ ] Is it simple?
 - [ ] Is it tested?
 - [ ] Is it documented?


s1113950:
UPDATE: tests pass if they run in a certain order, which is bad. Once that's fixed I'll remove the `wip` label here.

s1113950:
We're gonna cap at 2 threads for now because that's the default on circleci, but look into https://circleci.com/docs/2.0/parallelism-faster-jobs/ if we wanna make things faster. 8 cores works fine on my mac atow.

s1113950:
note to self: last push broke every major e2e test. Official fail is unable to undeploy.

s1113950:
Now we have things like this:

```
E       AssertionError: Popen call failed:
E       b"Starting build ae1fb177f-hello-world:63d8dff2-79b4-4cbc-a7fe-68cfd02ffd86\n\n\x1b[31mmake: *** No rule to make target `build'.  Stop.\n\x1b[0m\n"
E
E       b'\r                                                                               \r\rBuilding ae1fb177f-hello-world / (Elapsed Time: 0:00:00)                       \r                                                                               \r\rBuilding ae1fb177f-hello-world | (Elapsed Time: 0:00:00)                       \n'
``` that dumps the `stdout` and `stderr` if a popen call fails in a test. This coupled with #265 will make debugging quite easier! I needed to add this now because debugging threaded tests is impossible with no output

**Comment by s1113950** [comment]
UPDATE: tests run in about half the time total. We can still improve things by having more things preinstalled rather than installing on every run. Also, I think we still have the base resource amounts. I submitted a request to circleci to get `xlarge` machines like what we have with `NervanaSystems` repos.

**Comment by s1113950** [comment]
This is ready, but going to rip out things from it into other PRs and then it will be ready for review.

File: Makefile [hunk 1]
```diff
@@ -21,6 +21,28 @@ PY ?= $(shell python --version 2>&1  | cut -c8)
...
+ifeq ($(OS), Darwin)
+	NUMBER_OF_THREADS := $(shell sysctl -n hw.ncpu)
+else
+	NUMBER_OF_THREADS := $(shell nproc --all)
+endif
+
+# we want to leave a core per job for docker commands and things
+# but we always want at least 1 thread
+NUMBER_OF_THREADS_E2E := `expr $(NUMBER_OF_THREADS) / 2`
+ifeq ($(shell test $(NUMBER_OF_THREADS) -lt 1; echo $$?),0)
```
ashahba: I think this should be:
```sh
ifeq ($(shell test $(NUMBER_OF_THREADS) -le 1; echo $$?),0)
```

Makefile
```diff
@@ -21,6 +21,28 @@ PY ?= $(shell python --version 2>&1  | cut -c8)
...
+	NUMBER_OF_THREADS_E2E := 1
+endif
+
+# we also only have like 13 e2e tests; we don't need to raise this for awhile
+ifeq ($(shell test $(NUMBER_OF_THREADS) -gt 12; echo $$?),0)
+	NUMBER_OF_THREADS_E2E = 12
```
ashahba: I recommend defining another environment variable like:
```sh
MAX_NUMBER_OF_THREADS ?= 12
```
and then do:
```sh
ifeq ($(shell test $(NUMBER_OF_THREADS) -gt $MAX_NUMBER_OF_THREADS; echo $$?),0)
```

Makefile
```diff
@@ -129,11 +155,13 @@ test-e2e-setup: env-up
 	docker-compose exec test bash -c "/usr/share/mlt/scripts/ksync_install.sh"
 	docker-compose exec test bash -c "ln -sf /root/.ksync/bin/ksync /usr/local/bin/ksync"
 
-test-e2e: test-e2e-setup
-	docker-compose exec test env MLT_REGISTRY=gcr.io/intelai-mlt tox -e py${PY}-e2e
+test-e2e-circleci: test-e2e-setup-circleci
+	docker-compose exec test env MLT_REGISTRY=gcr.io/intelai-mlt env NUMBER_OF_THREADS=8 tox -e py${PY}-e2e
```
ashahba: This also might be defined like:
```sh
CI_MAX_NUMBER_OF_THREADS: 8
```
Near where you have the `xlarge` for docker file defined.
and then using that variable instead here and also move your comments up there so everything would be near each other and easy to maintain 🙂

Makefile
```diff
@@ -129,11 +155,13 @@ test-e2e-setup: env-up
...
 
-test-e2e-all: test-e2e-setup
-	docker-compose exec test env MLT_REGISTRY=gcr.io/intelai-mlt tox -e py2-e2e -e py3-e2e
+# NOTE: if we're on a circleci machine, for now we've got 8 cores so we'll cap at 8
+# check for core count isn't valid because we have a subset of cores available on a large host node
+test-e2e-all-circleci: test-e2e-setup-circleci
+	docker-compose exec test env MLT_REGISTRY=gcr.io/intelai-mlt env NUMBER_OF_THREADS=8 tox -e py2-e2e -e py3-e2e
```
ashahba: same as previous comment

tests/conftest.py
```diff
@@ -80,3 +82,12 @@ def wrapper(path, mock=None):
         return m
 
     return wrapper
+
+
+# each test that calls `mlt init` will create inside of this temp dir
```
ashahba: `create app directory` maybe?

Makefile
s1113950: Ahh! A case I missed. If `NUMBER_OF_THREADS` is 2, then we need to set E2E threads as 1 as well. Nice catch!
My initial case was if there's only 1 core available, then this resolves to `0`, and we want a limit of 1 thread:
```
robertso@robertso-mac02:~/Nervana/mlt$ BA=`expr 1 / 2`
robertso@robertso-mac02:~/Nervana/mlt$ echo $BA
0
```
The goal for `NUMBER_OF_THREADS_E2E` is this:
`1 <= NUMBER_OF_THREADS <= (TOTAL_THREADS ON SYSTEM / 2) <= 12`
(because if we go over the number of tests we have we don't see improvement). 
It's just difficult to do that in `make` :/

Makefile
s1113950: I like that too! It's cleaner.

Makefile
s1113950: 🤔 I think I'll use the `MAX_NUMBER_OF_THREADS_E2E` you mentioned earlier and save having to add yet another env var 😂 Lemme know what you think after it's done

PR #281, opened 2018-06-07
Add features.md for feature tracking

### LINKED ISSUE #173: Add features.md for feature tracking
In order to validate each feature against a user and track it's usefulness and maturity, we would like to have an overview of all features we have built or are planning to build

dmsuehir:
fixes #173

Added a `features.md` document with current features, as well as upcoming features.

## Reviewer Checklist

 - [ ] Do you understand it?
 - [ ] Is it correct?
 - [ ] Is it safe?
 - [ ] Is it robust?
 - [ ] Is it simple?
 - [ ] Is it tested?
 - [ ] Is it documented?


docs/features.md
```diff
@@ -0,0 +1,192 @@
+# MLT Features
```
nqn: Some of the reasoning behind the features file, would be to track the maturity of each feature as well. If it just landed, it being in alpha (or if it just not hardened yet).

nqn: And maybe we should add a note where we add flags, that new flags should be documented here?

docs/features.md
```diff
@@ -0,0 +1,192 @@
...
+|---------------------|-------------|
+| `<name>` | Name of the configuration parameter to remove. |
+
+### mlt build
+
+```
+  mlt build [--watch]
+```
+
+This command builds a local image for the current project directory.
```
nqn: Maybe add the detail about the timer which is started when a file change is detected.

dmsuehir: @nqn  I don't know which timer you mean...  I'm looking at the code and there's that sleep (https://github.com/IntelAI/mlt/blob/master/mlt/commands/build.py#L97) - is that what you're referring to?  I just tried it out and it seems like file changes are detected and the re-build starts as soon as I save the file.

PR #285, opened 2018-06-08
Template yamls validation against schema

### LINKED ISSUE #145: Validate yaml templates
When we get a template submitted to us, we should make sure it's a valid template. See what kube-controllers-go did here: https://github.com/NervanaSystems/kube-controllers-python/blob/master/crdlib/custom_resource_definition.py#L75

Also here is a good example:
https://stackoverflow.com/questions/3262569/validating-a-yaml-document-in-python

jitendra42:
fixes #145 

## Reviewer Checklist

 - [ ] Do you understand it?
 - [ ] Is it correct?
 - [ ] Is it safe?
 - [ ] Is it robust?
 - [ ] Is it simple?
 - [ ] Is it tested?
 - [ ] Is it documented?


mlt/commands/deploy.py
```diff
@@ -40,6 +41,9 @@ def __init__(self, args):
         build_helpers.verify_build(self.args)
 
     def action(self):
+        for template_yaml in glob.glob("k8s-templates/*.yaml"):
```
s1113950: Since this is the same for both deploy and build, what if it were just called `schema.validate()` and then move the logic into that?

PR #291, opened 2018-06-12
MTL logs improvement.

karthikvadla:
This way of reading logs is much faster compared to reading line by line.

## Reviewer Checklist

 - [x] Do you understand it?
 - [x] Is it correct?
 - [x] Is it safe?
 - [x] Is it robust?
 - [x] Is it simple?
 - [x] Is it tested?
 - [x] Is it documented?


mlt/utils/log_helpers.py
```diff
@@ -66,28 +68,26 @@ def _get_logs(prefix, since, namespace):
...
-        else:
-            print("Exception: {}".format(ex))
-        sys.exit()
+
+    # TODO: remove shell=True. and make log_cmd as List.
+    logs = process_helpers.run_popen(log_cmd,
+                                     shell=True,
+                                     stdout=True,
+                                     stderr=subprocess.PIPE)
+    if logs.poll() != 0:
```
s1113950:
```
Popen.poll()
Check if child process has terminated. Set and return returncode attribute. Otherwise, returns None.
```
So this if statement is `if it's still going (None) or errored`. What if it finished successfully in the time it takes to get to this `if` check? Then we won't get logs? 🤔 wouldn't we always want to `.communicate()` to see what happened?

karthikvadla: As far as i understand. `Popen.poll()` is `None` while process running. 
and once process finishes it returns the exit code `0`-success or `1`-errored. In either case it enters the block.

dmsuehir: I think Steven is asking what if it finishes so fast that when we call `Popen.poll()` and it returns 0 (success), then we skip this block and don't get logs?

PR #301, opened 2018-06-16
Updated tf-dist-mnist template to refactor data.py and model.py

tonyreina:
Updates the tf-dist-mnist template so that main.py is refactored into main.py, data.py, and model.py.  Data scientists can update model.py and data.py with their own model and data loaders. Also main.py uses monitored training session.

NOTE: I'm not sure how to handle the checkpoints directory listed in settings.py. It needs to be a directory common to all workers. So KVC or something.

## Reviewer Checklist

 - [ ] Do you understand it?
 - [ ] Is it correct?
 - [ ] Is it safe?
 - [ ] Is it robust?
 - [ ] Is it simple?
 - [ ] Is it tested?
 - [ ] Is it documented?


dmsuehir:
@tonyreina  it looks like this model takes a lot longer to train.  I ran it on the dls-us cluster, and it seemed to take about 15 minutes.  Is that expected?  That could be why the test is failing.  The end-to-end tests try to deploy every template and we expect the pods to go from `Running` to `Successful`, and I think that check times out after about 8 minutes.  Can we reduce the number of epochs or something to make this template complete faster?

s1113950:
@tonyreina we're also restricted in Circleci for every command being < 10 min in length total

tonyreina:
Yes. Feel free to reduce the number of training steps. It should be in the hook part of the code. (StopAtHook)

-Tony

PR #303, opened 2018-06-16
Fix the 'make lint' so that all Python files owned by repo are scanned.

### LINKED ISSUE #302: make lint needs to be enhanced to properly look for all files within repo and check proper code styles.
Currently we only scan `mlt` folder and this needs to be fixes so that any file making it to the repo is properly formatted.

ashahba:
fixes #302 

## Reviewer Checklist

 - [ ] Do you understand it?
 - [ ] Is it correct?
 - [ ] Is it safe?
 - [ ] Is it robust?
 - [ ] Is it simple?
 - [ ] Is it tested?
 - [ ] Is it documented?


ashahba:
This PR still ignores the rules [C901](https://lintlyci.github.io/Flake8Rules/rules/C901.html) and [F481](https://lintlyci.github.io/Flake8Rules/rules/F841.html).
Deciding about these 2 rules are deferred to the second round of the `lint` fixing specially after more testing.

s1113950:
Why would we want to keep F481? Variables assigned but are never used would never be a benefit afaict 🤔 If we need an `import blaaa` line somewhere for something to work we can `# noqa` it and it won't be linted :)

examples/distributed_unet/mlt_app/test_dist.py
```diff
@@ -575,9 +575,9 @@ def export_model(sess, input_tensor, output_tensor):
     tf.logging.info("Export model")
     # To view pb model file:  saved_model_cli show --dir saved_model --all
     sess.graph._unsafe_unfinalize()
-    import shutil
     MODEL_DIR = os.path.join(CHECKPOINT_DIRECTORY, "saved_model")
     tf.logging.info("model directory: {}".format(MODEL_DIR))
+    # import shutil
```
s1113950: old comments?

tests/unit/commands/test_logs.py
```diff
@@ -74,21 +74,21 @@ def test_logs_get_logs(json_mock, open_mock, verify_init, sleep_mock,
...
+    logs_command = LogsCommand({'logs': True, '--since': '1m', '--retries': 5})
     logs_command.config = {'name': 'app', 'namespace': 'namespace'}
 
     check_for_pods_readiness_mock.return_value = True
     process_helpers.return_value.poll.return_value = 0
     process_helpers.return_value.communicate.return_value = ('', '')
     with catch_stdout() as caught_output:
         logs_command.action()
-        output = caught_output.getvalue()
+        caught_output.getvalue()
```
s1113950: We should fix this test; it doesn't look like it's testing anything

tox.ini
```diff
@@ -23,6 +23,15 @@
 envlist = py{2,3}-{venv,lint,unit,e2e,coverage,dev}
 skip_missing_interpreters = true
 
+[flake8]
+ignore = C901, F841
```
s1113950: I'd not ignore F841; see general comment on PR about this

examples/distributed_unet/mlt_app/test_dist.py
ashahba: Yes this is just to keep these 2 commented lines around, in case at some point we want them back. I have no preference on keeping or deleting them.
But if we want to delete both lines `580` and `581` should be deleted same if we decide to keep.

```python
    # import shutil                                                             
    # shutil.rmtree(MODEL_DIR, ignore_errors=True)  # Remove old saved model 
```

tox.ini
ashahba: I agree with you.
Let me take a stab at fixing the failures when I remove it from `ignore` list.

tests/unit/commands/test_logs.py
ashahba: Good catch.

ashahba:
@s1113950 this should be good now.

tests/unit/commands/test_logs.py
```diff
@@ -233,6 +236,6 @@ def test_logs_keyboardinterrupt(json_mock, open_mock, verify_init, sleep_mock,
 
     check_for_pods_readiness_mock.return_value = True
     process_helpers.side_effect = KeyboardInterrupt
-    with catch_stdout() as caught_output:
+    with catch_stdout() as _:  # noqa
```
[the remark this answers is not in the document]
ashahba: The test is good as long as we have the correct exception raised which in this case is `SystemExit` 🙂

PR #312, opened 2018-06-28
Adds support for picking pod to connect to interactively

### LINKED ISSUE #141: Support multiple pod interactive mode
Right now `--interactive` requires either 1 template to deploy or you specifying which template you want. We should allow for N of them.

s1113950:
fixes #141 

## Reviewer Checklist

 - [x] Do you understand it?
 - [x] Is it correct?
 - [x] Is it safe?
 - [x] Is it robust?
 - [x] Is it simple?
 - [x] Is it tested?
 - [x] Is it documented?


s1113950:
Now users see this:
```
robertso@sylviasr-mobl1:/tmp/py-dist$ mlt deploy --no-push --interactive
Skipping image push
Deploying gcr.io/constant-cubist-173123/py-dist:1ac7ee80-d20e-4f08-b709-1d0a3f91fa71

$ kubectl get --namespace=robertso all
or
$ mlt status

More than one container created or custom deployment chosen.
Call `kubectl exec -it {pod_name_here} --namespace=robertso /bin/bash` on a `Running` pod NAME below.
If no pods are running yet, run `mlt status` occasionally, or `watch -n1 mlt status` to watch until pods are `Running`.

Waiting for pods...
NAME                                                           READY     STATUS    RESTARTS   AGE
py-dist-5c6aac01-d83b-45e6-a5e9-ccb3e513-master-1l5f-0-vcl9d   2/2       Running   0          <invalid>
py-dist-5c6aac01-d83b-45e6-a5e9-ccb3e513-worker-1l5f-1-oyuf3   1/1       Running   0          <invalid>
py-dist-5c6aac01-d83b-45e6-a5e9-ccb3e513-worker-1l5f-2-dysdp   1/1       Running   0          <invalid>
```

mlt/commands/deploy.py
```diff
@@ -163,14 +175,38 @@ def _deploy_new_container(self):
...
+        self._update_app_run_id(app_run_id)
+        print("\nInspect created objects by running:\n"
+              "$ kubectl get --namespace={} all\n"
+              "or \n$ mlt status\n".format(self.namespace))
+
+        if self.args["--interactive"] and not self._replicas_found \
+                and self._total_containers == 1:
+            self._exec_into_pod(self._get_most_recent_podname())
+        else:
+            # TODO: decide if we always want to have custom deployments not be
```
WafaaT: Not all `custom-deploy-templates` won't have a yaml file, some changes will be added here in this PR: https://github.com/IntelAI/mlt/pull/315, I do custom deploy for the tensorbaord template but still use  k8s deployments using `k8s-templates/. .yaml`.

mlt/commands/deploy.py
```diff
@@ -163,14 +175,38 @@ def _deploy_new_container(self):
...
+            # able to be auto-exec-ed into. Reasoning behind this right now is
+            # that we read the yaml spec to determine how many pods there'll be
+            # for custom deployment we could count the number of pods maybe?
+            print("More than one container created or custom deployment chosen"
```
WafaaT: the same here regarding custom deploy!

mlt/commands/deploy.py
s1113950: @WafaaT do you think it makes sense for me to hold off on this PR until #315 goes in? Then I can use tensorboard as an example of a custom deploy with a yaml as well? wdyt?

mlt/commands/deploy.py
```diff
@@ -179,44 +215,35 @@ def _default_deploy(self, app_name, app_run_id, remote_container_name):
...
-                # if we never hit this point, throw error at end of deployment
-                # process because we wanted an interactive deployment
-                # but nothing matched criteria
-                self.interactive_deployment_found = True
-        return interactive, data
+        # TODO: we thought about not calling _patch_template_spec if there's
+        # not a `k8s` folder present. Current implementation is we just call
+        # `make deploy` and that's it, so we'll have to decide that we want
+        # to still try and `_patch_template_spec` even though we already called
+        # `make deploy` if there's a k8s folder present
```
WafaaT: I think that `make deploy` calls the default deploy method which has `k8s-templates and k8s` directories. otherwise, it would be a `custom deploy` that probably has `k8s-templates and k8s` or not. are there cases with default deploy which as no `k8s`??
also Dina, Karthik and I  agreed to add a condition on the `k8s-templates` directory existence for custom-deploy case. I'm working on that in #315. what do you think?

resources/wait-port.sh
```diff
@@ -17,6 +17,8 @@
 # SPDX-License-Identifier: EPL-2.0
 #
 
+# NOTE: UNUSED BUT KEEPING AROUND FOR FUTURE LOCAL CLUSTER CREATION
```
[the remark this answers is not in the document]
s1113950: It was called here: https://github.com/IntelAI/mlt/blob/v0.1.0-alpha1/Makefile#L70

mlt/commands/deploy.py
WafaaT: @s1113950 I can update you in 2 hrs, I'm adding some changes to `_deploy_new_container` if the tests passed then we can have #315 in first, Is that OK? what do you think?

s1113950: Sounds good! I think it makes sense to put #315 in first

mlt/main.py
```diff
@@ -76,20 +76,22 @@
...
-                            and then drops user into `kubectl exec` shell.
-                            Adds a `debug=true` label for easy discovery
-                            later. If you have more than 1 template yaml,
-                            specify which file you'd like to deploy
-                            interactively as the `kube_spec`. `kube_spec` is
-                            only used with this flag.
+  --interactive             Rewrites all container commands to infinite sleep,
+                            and then drops user into `kubectl exec` shell if
+                            there's only one container deployed. Otherwise,
+                            outputs helpful text to enable you to connect to
```
ashahba: Too many `to`'s 🙂 can we say: `outputs helpful text to help you connect to`

tests/e2e/test_deploy_flow.py
```diff
@@ -36,6 +36,7 @@
                              os.path.join('mlt-templates', x)),
                              os.listdir('mlt-templates')))
 def test_deploying_templates(template):
+    """tests every folder in our mlt-templates dir"""
```
ashahba: `every folder` or `every template`?

tests/e2e/test_deploy_flow.py
```diff
@@ -83,10 +84,12 @@ def test_no_push_deploy():
             commands.teardown()
 
 
-def test_interactive_deploy():
+@pytest.mark.parametrize('template', ['hello-world', 'tf-distributed'])
+def test_interactive_deploy(template):
+    """tests 1 pod made and template with many pods made"""
```
ashahba: `tests templates with either a single pod or multiple pods`

mlt/commands/deploy.py
s1113950: New PR: https://github.com/IntelAI/mlt/pull/328

tests/e2e/test_deploy_flow.py
s1113950: It's technically both of those things but I'll clarify with `every template`

tests/e2e/test_deploy_flow.py
s1113950: :lol: yeah I'll change this

s1113950:
We don't need to have the `if os.path.isdir('k8s-templates')` logic because of this:
```
In [10]: for path, dirs, files in os.walk('banana'):
    ...:     print('baaaaaaaa')
    ...:

In [11]:
``` 
if there's no dir there, there's no error and nothing happens :)
Given that `default_deploy` _only_ calls `for path, dirs, filenames in os.walk("k8s-templates"):` to do the default deploy logic now, I'm gonna tweak the deploy logic a bit

mlt/commands/deploy.py
```diff
@@ -163,14 +175,38 @@ def _deploy_new_container(self):
...
+            print("More than one container created or custom deployment chosen"
+                  ".\nCall `kubectl exec -it {{pod_name_here}} "
+                  "--namespace={} /bin/bash` on a `Running` pod NAME "
+                  "below.\nIf no pods are running yet, run `mlt status` "
+                  "occasionally, or `watch -n1 mlt status` to watch until "
+                  "pods are `Running`.\n".format(self.namespace))
+            # TODO: how do we we wanna do this? Sometimes it takes a bit for
+            # crds to launch stuff
+            print('Waiting for pods...')
+            time.sleep(3)
```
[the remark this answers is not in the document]
s1113950: What about not waiting at all? I have a message for user sto do `watch -n1 mlt status` if they want to wait for a pod to be `Running`; does it make sense you think to add logic to manage that when there are various ways users can do this? Or even just go for a break, come back, and then exec into a pod? WDYT?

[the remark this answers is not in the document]
ashahba: Most end users should be intelligent enough to know if the Pod is still being created, as long as robots/scripts don't blindly try to just ahead to the next command before we actually have the pods running, then we shouldn't need to write our custom logic to check for Pod's readiness.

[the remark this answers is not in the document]
s1113950: Sounds good!

mlt/commands/deploy.py
```diff
@@ -153,33 +153,54 @@ def _deploy_new_container(self):
...
+        if we find > 1 container regardless of replica, same logic applies
+        """
+        self._replicas_found = False
+        self._total_containers = 0
+
+        # deploy our normal template sub logic, then if `deploy` in Makefile
+        # add whatever custom stuff is desired
+        self._default_deploy(app_name=app_name,
+                             app_run_id=app_run_id,
+                             remote_container_name=remote_container_name)
```
WafaaT: Here we call default deploy first for all templates, what if it doesn't have `k8s-templates`, does it error out?

mlt/commands/deploy.py
```diff
@@ -241,64 +247,71 @@ def _get_most_recent_podname(self):
...
+        # TODO: confirm this, it appears that the yaml always makes
+        # `containers:` a list of lists
+        for location_list in self.containers_locations:
+            for location in location_list:
+                self._total_containers += 1
+                # https://kubernetes.io/docs/tasks/inject-data-application/
+                # define-command-argument-container/#notes
+                # this will override both ENTRYPOINT and Cmd to ensure sleep
+                location.update({
+                    'command': ["/bin/bash"],
```
ashahba: What if `bash` is not provided by `container` and we only have `/bin/sh` to work with?

mlt/main.py
```diff
@@ -76,20 +76,22 @@
...
+                            your running container.
+                            Adds a `debug=true` label for easy discovery later.
   --retries=<retries>       Number of times to retry while waiting for pods to
                             be running. Waits 1 second between retrying. Used
                             with interactive deploy and logs.
                             [default: 120]
   --since=<duration>        Returns logs newer than a relative
                             duration like 10s, 1m, or 2h [default: 1m].
   --logs                    Tail logs after deploying [default: False]
+  podname                   Optionally name a container to connect to. If this
```
ashahba: I don't thik this is aligned correctly 🤔

mlt/commands/deploy.py
```diff
@@ -153,33 +153,54 @@ def _deploy_new_container(self):
...
                                 app_run_id=app_run_id,
                                 remote_container_name=remote_container_name)
+
+        self._update_app_run_id(app_run_id)
+        print("\nInspect created objects by running:\n"
+              "$ kubectl get --namespace={} all\n"
+              "or \n$ mlt status\n".format(self.namespace))
+
+        if self.args["--interactive"] and not self._replicas_found \
+                and self._total_containers == 1:
```
ashahba: With these restrictions we almost never shell into the container automatically, as it is the case with only single Pod, single container deployments like `hello-world` application.
One may ask should we even bother with shelling into the pod and keep everything consistent 🤔 and only put the Pod in `sleep infinity & wait` mode? WDYT?

mlt/commands/deploy.py
s1113950: It simplifies the logic a bit (no need for if/else statement). If there's no dir, then nothing happens:
```
In [10]: for path, dirs, files in os.walk('banana'):
    ...:     print('baaaaaaaa')
    ...:

In [11]:
```
so then `default_deploy` will do nothing in this case.
If people don't like this then I can revert it, but it looked cleaner to me 🤷‍♂️

mlt/commands/deploy.py
s1113950: Ahh good point! 🤔 We had `bash` as the shell to use earlier as well. What do you think of another ticket to identify what shell the user has and run that one instead? It would mean checking what the `entrypoint` or `cmd` inside the `dockerfile` says and parsing the shell out of it. @ashahba

mlt/main.py
s1113950: Turns out I didn't end up using it! We decided to change the logic for this. So I'll just delete it

mlt/commands/deploy.py
s1113950: That's a good point! I left it like this because we had a meeting with Tyler where he said he liked the auto-shelling so I left it in :) if in the future it's not really used we can rip it out?

mlt/commands/deploy.py
ashahba: I think we can just add a comment and tag it as a `TODO:` item for future since as far as I know all the `Dockerfile`'s we have support `bash` and we haven't had an issue yet.
So just a comment so that we can find it easily if we want to do that work at some point should be fine for now.

mlt/commands/deploy.py
ashahba: Ok, so in that case it should be good as it is.

PR #321, opened 2018-07-02
Add `debug_on_fail` option to templates

### LINKED ISSUE #305: Add `debug on fail` pdb debugging capability to existing templates
This request comes from the [proposal](https://github.com/NervanaSystems/dls-features/issues/79) for dls-feature [#79](https://github.com/NervanaSystems/dls-features/issues/79).  

Add a debug wrapper, which catches any exceptions when running the code, and sets pdb breakpoint.  This allows the user to be able to attach to the pod, inspect variables, etc for debugging.  There should be an environment variable flag to enable (`DEBUG_ON_FAIL`) and documentation that explains how it works.  Also, note that when the failure occurs, since we are catching the exception and just sitting at a breakpoint, the Kubernetes pod doesn't show as errored (it still shows as running), so logs must be monitored to see if an error occurs.

dmsuehir:
fixes #305 

Adds support for using the `kubernetes_debug_wrapper.py` from templates to catch unhandled python exceptions and then debug using pdb.  See the docs/template_developers.md to see how to use the debug wrapper when writing a new template.

This PR enables the debug wrapper for `hello-world` and `tf-dist-mnist`.

Also, updated the circleci test cluster kubernetes version to `1.10.5-gke.3` to get the fix for the issue where backoff limit doesn't work with 1.10.0, but was fixed in 1.10.5 (see https://github.com/NervanaSystems/dls-infrastructure/issues/26).

## Reviewer Checklist

 - [ ] Do you understand it?
 - [ ] Is it correct?
 - [ ] Is it safe?
 - [ ] Is it robust?
 - [ ] Is it simple?
 - [ ] Is it tested?
 - [ ] Is it documented?


mlt/commands/init.py
```diff
@@ -89,9 +89,19 @@ def action(self):
                         sys.exit(1)
 
                 data = self._build_mlt_json(template_params, template_git_sha)
+
```
s1113950: on `init` do we also want to edit the template yaml to put the `tty` and `stdin` values there as well as the necessary DEBUG_ON_FAIL stuff? It's the same for all templates right? If a user runs `mlt deploy --debug` or something we could enable that for them rather than hardcoding it :)

tests/e2e/test_deploy_flow.py
```diff
@@ -111,3 +113,34 @@ def test_watch_build_and_deploy_no_push():
...
+            fh.write("raise ValueError()\n")
+
+        commands.build()
+        try:
+            # Pod should still be running instead of going to 'Succeeded'
+            # since debug is enabled and we inserted an error
+            commands.deploy(pod_stays_running=True)
+
+            # Pod should still be running to allow user to debug
+            time.sleep(5)
```
s1113950: are we sure 5 sec will always be enough time?

tests/test_utils/e2e_commands.py
```diff
@@ -306,17 +312,20 @@ def _verify_pod_success(self, interactive_deploy, sync_deploy):
...
-        if not interactive_deploy and not sync_deploy:
+        if not pod_stays_running:
             # since new pods could come up, we might find another 'Pending' pod
             while pod_status == 'Running' or pod_status == 'Pending':
                 time.sleep(1)
                 pod_status = self._grab_latest_pod_or_tfjob()
                 if time.time() - start >= 480:
                     break
-            assert pod_status == 'Succeeded'
+            assert pod_status == 'Succeeded', \
```
s1113950: :woo:

mlt/commands/init.py
dmsuehir: It won't necessarily work for all templates, for example it won't work for tensorboard.

mlt-templates/tf-dist-mnist/k8s-templates/tfjob.yaml
```diff
@@ -13,6 +13,11 @@ spec:
           containers:
             - image: $image
               name: tensorflow
+              tty: true
```
WafaaT: how about adding a comment that these parameters `tty` and `stdin` are added just to enable the debugger?

tests/test_utils/e2e_commands.py
```diff
@@ -237,7 +237,19 @@ def deploy(self, no_push=False, interactive=False,
         if self.template == "tensorboard":
             pod_stays_running = True
 
-        self._verify_pod_success(pod_stays_running)
+        self._verify_pod_success(interactive, pod_stays_running)
```
WafaaT: I think the e2e test fails here, you modified `_verify_pod_success` to take self and `pod_stays_running` right?

dmsuehir: Yeah, I am fixing it

mlt/commands/init.py
dmsuehir: I'm trying to do the first thing you mentioned with editing the yaml from the init instead of adding it to the yaml, but it looks like the python library (I'm using `yaml.dump`) does weird things with some of the values.  For example, if I just do `yaml.load()` and `yaml.dump` without editing anything in the dictionary, it looks something like this:
```
apiVersion: batch/v1
kind: Job
metadata: {name: $app-$run}
spec:
  backoffLimit: 4
  template:
    spec:
      containers:
      - env:
        - {name: GREETING, value: $greeting}
        image: $image
        name: $app
      restartPolicy: Never
```
Where there are `{` and `}` for some of the values, which weren't there before.

s1113950: I ran into that before 🤔 Check out `_patch_template_spec()` function for how I did it; I solved it with a `json.dumps(data)` instead :trollface: That should work for our case too because I'm modifying the template spec and K8s can take a json blob too :) @WafaaT

Makefile
```diff
@@ -175,17 +175,20 @@ test-e2e-all-circleci: test-e2e-setup-circleci
 # if you'd like to use something other than localhost:5000, also set
 # MLT_REGISTRY env var and that'll be respected by tox
 test-e2e-no-docker-setup:
+	[ "$(command -v kubectl version)" ] && echo "ba" || echo "wa"
```
ashahba: I don't think we still need this.

s1113950: oh shoot my bad! leftover debugging

Makefile
```diff
@@ -175,17 +175,20 @@ test-e2e-all-circleci: test-e2e-setup-circleci
...
 	@if [ "$(command -v kubectl version)" ]; then \
         [ `kubectl get crd | grep -E 'tfjobs\.kubeflow\.org|pytorchjobs\.kubeflow\.org' -c` -eq "2" ] || \
```
ashahba: I think this line maybe contributing to the issue, but I can't quite figure it out.

s1113950:
NOTE TO SELF: remember to put back in `ksync` after https://github.com/vapor-ware/ksync/issues/219 is resolved for us

mlt-templates/hello-world/k8s-templates/job.yaml
```diff
@@ -16,3 +16,4 @@ spec:
         - name: GREETING
           value: $greeting
       restartPolicy: Never
+  backoffLimit: 4
```
karthikvadla: is it like retry?

mlt-templates/hello-world/parameters.json
```diff
@@ -1,5 +1,6 @@
 {
   "template_parameters" : [
-       { "name": "greeting", "value": "Hello" }
+    { "name": "greeting", "value": "Hello" },
+    { "name": "debug_on_fail", "value": "False" }
```
karthikvadla: We have to set ENTRYPOINT with k8_wrapper only when it is TRUE?

mlt-templates/hello-world/k8s-templates/job.yaml
s1113950: I didn't add this part (@dmsuehir did originally), but yeah it specifies the number of retries before considering a Job as failed. By default it is 6. Not sure why it's been changed to 4 but it didn't hurt anything so I left it alone. If you think it should be different lemme know!

mlt-templates/hello-world/parameters.json
s1113950: I think we're setting it always because then user could flip it on and off without editing their Dockerfile every time

karthikvadla:
yeah looks good to me.

s1113950:
Note to me: I meant "2 min count" above, not 3 min

s1113950:
Discovered I can get this to work when I run it this way:
```
EXTRA_ARGS=`$MLT_REGISTRY_AUTH_COMMAND` TESTFILES="tests/e2e/test_deploy_flow.py::TestDeployFlow::test_deploy_check_logs" TESTOPTS="-s" PY=2 make test-e2e-no-docker
```
Need to figure out why `pytest` capturing is causing issues specifically with our `python2` `kubetail` test.
It should be something with this PR because checking logs worked before. 🤔

s1113950:
Tests pass now but hang at `100%` 🤔 there's a process hanging around I think. Side effect of all the `subprocess` calls we do probably...

s1113950:
From here: https://circleci.com/gh/IntelAI/mlt/1897?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link discovered it's this test that hangs: `tests/e2e/test_deploy_flow.py::TestDeployFlow::test_interactive_deploy[hello-world]`

PR #328, opened 2018-07-05
Wafaa/add custom deploy tensorboard template

### LINKED ISSUE #227: Launch Tensorboard service in a local browser using MLT
Having Tensorboard service deployed and running in Kubernetes, `mlt` existing or new command is required to do (automate) the following:
- Get the TB service external IP address.
- Form the HTTP request to access TB such as `http:\\<external_ip_address>:6006`.
- Access the user local browser and launch TB.  

This is required in phase 2 for the dls-feature proposal https://github.com/NervanaSystems/dls-features/pull/121

WafaaT:
fixes #227

## Reviewer Checklist

 - [ ] Do you understand it?
 - [ ] Is it correct?
 - [ ] Is it safe?
 - [ ] Is it robust?
 - [ ] Is it simple?
 - [ ] Is it tested?
 - [ ] Is it documented?

### DIFF
--- a/mlt-templates/tensorboard/README.md
+++ b/mlt-templates/tensorboard/README.md
@@ -67,12 +67,8 @@ In this example, the user need to modify k8s-templates/tbjob.yaml to add the pre
 ```
 
 4. Deploy TensorBoard service.
-`mlt deploy` 
-
-5. Get the service `EXTERNAL-IP`.
-`mlt status`
-
-6. Open TensorBoard: launch browser with `http://<EXTERNAL-IP>:6006`.
+`mlt deploy`,
+which deploys the TensorBoard service and launches the local browser with `http://<EXTERNAL-IP>:6006`.
 
 7. Finally, when done, a user can delete the running TensorBoard service
 `mlt undeploy`.
--- a/mlt-templates/tensorboard/README.md
+++ b/mlt-templates/tensorboard/README.md
@@ -9,7 +9,7 @@ The command below shows how to update the log directory value in `k8s-templates/
 
 TensorBoard template works with different storage sources, and it's the user responsibility to allow the TensorBoard service to access the logs location.
 
-This template assumes that Ingress controller is enabled in the Kubernetes cluster. 
+This template assumes that Ingress controller is enabled in the Kubernetes cluster, and also that `jq` is installed. `jq` is used while launching TensorBoard to get the service IP address.
 
 
 
--- a/mlt-templates/tensorboard/deploy.sh
+++ b/mlt-templates/tensorboard/deploy.sh
@@ -21,21 +21,21 @@
 if [ ! -z "$( kubectl get service ${JOB_NAME} --namespace ${NAMESPACE} )" ]
 then
   ip_address='none'
-  # loop until the ip address is assigned to the service
-  while [ 1 ]
+  timer=0
+  # loop until the ip address is assigned to the service or the timer is done.
+  while (( timer < 120))
   do
-    ip_address=$( kubectl get service ${JOB_NAME} --namespace ${NAMESPACE} no -o json | jq -r '.items[0].status.loadBalancer.ingress[].ip' )
-    if [[ -z ${ip_address} ]]
+    ip_address=$( kubectl get service ${JOB_NAME} --namespace ${NAMESPACE} -o json | jq -r '.status.loadBalancer.ingress[]?.ip' )
+    if [[ ! -z ${ip_address} ]]
     then
-        echo "Reading the service address..."
-    else
         echo "Launching TensorBoard in your local browser..."
         # open TB in the default browser
         open http://${ip_address}:6006
         exit
     fi
+    timer=$[$timer+1]
     sleep 5
   done
 else
-  echo "Cannot open TensorBoard in your local browser, please try mlt status to check if the service is running."
+  echo "Cannot open TensorBoard in your local browser, please try 'mlt status' to check if the service is running."
 fi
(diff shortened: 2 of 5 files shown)


mlt-templates/tensorboard/deploy.sh
```diff
@@ -0,0 +1,41 @@
...
+
+# check if the service is up and running and get the TB service IP address
+
+if [ ! -z "$( kubectl get service ${JOB_NAME} --namespace ${NAMESPACE} )" ]
+then
+  ip_address='none'
+  # loop until the ip address is assigned to the service
+  while [ 1 ]
+  do
+    ip_address=$( kubectl get service ${JOB_NAME} --namespace ${NAMESPACE} no -o json | jq -r '.items[0].status.loadBalancer.ingress[].ip' )
```
WafaaT: Any other suggestions for reading the `External_IP` address?

ashahba: The only caution point here is, if `loadBalancer` is empty:
```
$ kubectl get svc hello-world no -o json | jq -r '.items[0].status'
Error from server (NotFound): services "no" not found
{
  "loadBalancer": {}
}
```

mlt-templates/tensorboard/deploy.sh
```diff
@@ -0,0 +1,41 @@
...
+# SPDX-License-Identifier: EPL-2.0
+#
+
+# check if the service is up and running and get the TB service IP address
+
+if [ ! -z "$( kubectl get service ${JOB_NAME} --namespace ${NAMESPACE} )" ]
+then
+  ip_address='none'
+  # loop until the ip address is assigned to the service
+  while [ 1 ]
```
s1113950: I would have a default timeout just in case the ip address is never assigned to the service, like 1 min or something? Just to prevent an infinite loop here.

mlt-templates/tensorboard/deploy.sh
```diff
@@ -0,0 +1,41 @@
...
+    else
+        echo "Launching TensorBoard in your local browser..."
+        # open TB in the default browser
+        open http://${ip_address}:6006
+        exit
+    fi
+    sleep 5
+  done
+else
+  echo "Cannot open TensorBoard in your local browser, please try mlt status to check if the service is running."
```
s1113950: I would add quotes around `mlt status`, like `'mlt status'` so it's more clear that's the command to run

mlt/commands/deploy.py
```diff
@@ -154,7 +154,16 @@ def _deploy_new_container(self):
         print("Deploying {}".format(remote_container_name))
         kubernetes_helpers.ensure_namespace_exists(self.namespace)
         app_run_id = str(uuid.uuid4())
+
         if files.is_custom("deploy:"):
+            # this checks if we use yaml files in K8s deployments in custom
+            # deploy case and do the template parameters substitutions.
+            if os.path.isdir("k8s-templates"):
```
s1113950: This is fine for now because my PR will refactor how default and custom deploys work!

mlt-templates/tensorboard/deploy.sh
ashahba: For this one and the case of `ingress` using `10` minutes for timeout should be good.

mlt-templates/tensorboard/deploy.sh
WafaaT: @ashahba any idea how to fix that?!

mlt/commands/deploy.py
karthikvadla: @WafaaT Can you please add unit test case for this.

mlt-templates/tensorboard/deploy.sh
ashahba: @WafaaT 
`$?` still returns `0` so we can rely on command exit code.

So I would do something more intelligent like this which doesn't throw errors:
```
$ ip_address=$( kubectl get service ${JOB_NAME} --namespace ${NAMESPACE} no -o json | jq -r '.items[0].status.loadBalancer.ingress[].ip' 2>&1 )
```
Now if I look at `$ip_address` before using it, I can tell if it's valid ip or not:
```
echo $ip_address 
jq: error (at <stdin>:9): Cannot iterate over null (null)
```

and take actions based on the value of `$ip_address`.

mlt-templates/tensorboard/deploy.sh
ashahba: And you can use something like this pattern:
```
n=0
   until [ $n -ge 120 ]
   do
     if <YOUR_COMMAND>; then break
      n=$[$n+1]
      sleep 5
   done
```
This will exit if the `YOUR_COMMAND` exit code `$?` is `0` 🙂

mlt/commands/deploy.py
WafaaT: @karthikvadla , can we skip adding the unit test now until @s1113950 refactor this code in PR #312 ?

mlt-templates/tensorboard/README.md
```diff
@@ -9,7 +9,7 @@ The command below shows how to update the log directory value in `k8s-templates/
 
 TensorBoard template works with different storage sources, and it's the user responsibility to allow the TensorBoard service to access the logs location.
 
-This template assumes that Ingress controller is enabled in the Kubernetes cluster. 
+This template assumes that Ingress controller is enabled in the Kubernetes cluster, and also that `jq` is installed. `jq` is used while launching TensorBoard to get the service IP address.
```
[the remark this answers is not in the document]
WafaaT: @s1113950 I can add it, but as it's not optional for the tensorboard template, can we leave this note here too?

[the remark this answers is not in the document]
s1113950: If you want to, sure! The `optional prerequisites` is more like, `certain commands won't work without these things installed` section

s1113950:
nvmd, having them together should be okay!

PR #338, opened 2018-07-10
Fix the Makefile to support running dev commands with specific version of Python, like PY=3 make dev-env

ashahba:
Now you can do things like:

```
$ make dev-env 
Creating Python2 editable install in virtualenv
GLOB sdist-make: /Users/ashahba/source_code/IntelAI/mlt/setup.py

$ PY=3 make dev-env 
Creating Python3 editable install in virtualenv
GLOB sdist-make: /Users/ashahba/source_code/IntelAI/mlt/setup.py
...
```

## Reviewer Checklist

 - [ ] Do you understand it?
 - [ ] Is it correct?
 - [ ] Is it safe?
 - [ ] Is it robust?
 - [ ] Is it simple?
 - [ ] Is it tested?
 - [ ] Is it documented?


Makefile
```diff
@@ -16,7 +16,8 @@
 #
 
 SHELL=bash
-PY := $(shell python --version 2>&1  | cut -c8)
+# Set PY_VER to 2 or 3 to test with Python2 or Python3 versions
+PY := $(shell python${PY_VER} --version 2>&1  | cut -c8)
```
s1113950: Not quite, if you want to override this though just do `PY ?= $(shell...`, which will take from the env if it's specified. No need to pass a different var in.

docs/developer_guide.md
```diff
@@ -9,6 +9,9 @@ To get set up, please run the following command:
 `make dev-env`
 
 This command will create an editable install of your `mlt` repository.
+To install specifically with `Python3` version, set `PY_VER` variable and run the command as follows:
```
s1113950: ```set `PY` variable and run...```

Makefile
ashahba: Thanks @s1113950 `?` was the part that I was missing and couldn't initially get it to work 🙂

PR #351, opened 2018-07-18
Ashahba/issue 121

ashahba:
Partially fixes #121 for developers.

Once this is merged we can do another PR to fix the issue for end users who use `mlt` binary as well.

## Reviewer Checklist

 - [ ] Do you understand it?
 - [ ] Is it correct?
 - [ ] Is it safe?
 - [ ] Is it robust?
 - [ ] Is it simple?
 - [ ] Is it tested?
 - [ ] Is it documented?


Makefile
```diff
@@ -15,11 +15,14 @@
...
-SHELL=bash
+SHELL := bash
+EXES := python tox docker docker-compose
+OK := $(foreach exec,$(EXES), $(if $(shell which $(exec)),,$(error "No $(exec) found in PATH")))
+
 # Set PY to 2 or 3 to test with Python2 or Python3 versions
 PY ?= $(shell python --version 2>&1  | cut -c8)
-VIRTUALENV_DIR=$(if $(subst 2,,$(PY)),.venv3,.venv)
-ACTIVATE="$(VIRTUALENV_DIR)/bin/activate"
+VIRTUALENV_DIR := $(if $(subst 2,,$(PY)),.venv3,.venv)
```
s1113950: What's wrong with lazy setting of these 2 vars? https://stackoverflow.com/a/448939 if they aren't used then we never have to initialize them

Dockerfile
```diff
@@ -40,13 +40,17 @@ RUN apt-get update && apt-get install -yq --no-install-recommends --fix-missing
...
 
+RUN pip install tox
 RUN curl -fsSL https://download.docker.com/linux/ubuntu/gpg | apt-key add -
 RUN add-apt-repository "deb [arch=amd64] https://download.docker.com/linux/ubuntu $(lsb_release -cs) stable"
 RUN apt-get update && apt-get install -yq --no-install-recommends --fix-missing \
     docker-ce && \
     apt-get clean && \
     rm -rf /var/lib/apt/lists/*
 
+RUN curl -L https://github.com/docker/compose/releases/download/1.21.2/docker-compose-$(uname -s)-$(uname -m) -o /usr/local/bin/docker-compose && \
```
s1113950: I'd say let's not put this in the dockerfile; it's used in circleci but from the host machine, not within `mlt`.

Makefile
```diff
@@ -172,8 +175,12 @@ test-e2e-all-circleci: test-e2e-setup-circleci
 # if you'd like to use something other than localhost:5000, also set
 # MLT_REGISTRY env var and that'll be respected by tox
 test-e2e-no-docker-setup:
-	@[ `kubectl get crd | grep -E 'tfjobs\.kubeflow\.org|pytorchjobs\.kubeflow\.org' -c` -eq "2" ] || \
-		GITHUB_TOKEN=${GITHUB_TOKEN} ./scripts/kubeflow_install.sh
+	@ if [ -x "$(command -v kubectl version)" ]; then \
+        [ `kubectl1 get crd | grep -E 'tfjobs\.kubeflow\.org|pytorchjobs\.kubeflow\.org' -c` -eq "2" ] || \
+		GITHUB_TOKEN=${GITHUB_TOKEN} ./scripts/kubeflow_install.sh; \
+	else \
+		echo "Please install kubectl"; \
```
s1113950: I'd throw an exit code after the echo call, otherwise the `no-docker-setup` will terminate "successfully"

Makefile
ashahba: Nothing wrong just consistency, but they can remain like before WDTY?

Dockerfile
ashahba: I agree, but if we remove it this line fails:
https://github.com/IntelAI/mlt/pull/351/files#diff-b67911656ef5d18c4ae36cb6741b7965R20
unless we remove the check for `docker-compose`.
That's why I asked if we have to remove the check for `docker-compose` since we only use it in `CircleCi`.

Makefile
s1113950: I'd keep them as before yeah. We've kept vars before in the past that expand lazy and it's helped us! 🤷‍♂️

Dockerfile
s1113950: oh! I see what you mean now. Yeah I'd remove it. The `make` call is `test-e2e-circleci` so people shouldn't be running that one :)

PR #352, opened 2018-07-18
Update mlt codebase to check for installation of required packages.

ashahba:
Second phase of fixes for #121 

## Reviewer Checklist

 - [ ] Do you understand it?
 - [ ] Is it correct?
 - [ ] Is it safe?
 - [ ] Is it robust?
 - [ ] Is it simple?
 - [ ] Is it tested?
 - [ ] Is it documented?


ashahba:
Looks like for commands like:
```
build_cmd = "CONTAINER_NAME={} GPUS={} make build".format(
            container_name, template_parameters.get('gpus', 0))
```
in [build.py](https://github.com/IntelAI/mlt/blob/master/mlt/commands/build.py#L58) I need to have a better way of identifying the command before validation and this is why the `e2e` tests are failing.

Dockerfile
```diff
@@ -48,6 +48,9 @@ RUN apt-get update && apt-get install -yq --no-install-recommends --fix-missing
     apt-get clean && \
     rm -rf /var/lib/apt/lists/*
 
+RUN curl -L https://github.com/docker/compose/releases/download/1.21.2/docker-compose-$(uname -s)-$(uname -m) -o /usr/local/bin/docker-compose && \
```
s1113950: Why do we need `docker-compose` in the dockerfile? It's not required to run MLT

mlt/utils/process_helpers.py
```diff
@@ -40,5 +42,17 @@ def run_popen(command, shell=False, stdout=PIPE, stderr=PIPE, cwd=None):
     with open(os.devnull, 'w') as quiet:
         stdout = quiet if stdout is False else stdout
         stderr = quiet if stderr is False else stderr
+        if not (isinstance(command, str) or isinstance(command, list)):
```
s1113950: Do we need these checks? There are 2 types of commands: 
1. Strings
2. Arrays
Strings are typically used if the command includes a `|` because you can't run that as an argument array.
In this case, users should run with `shell=True`.
I suppose we could auto set the `shell=True` if the command is a string, but we can't do what you're doing on line 48 because of the `pipe` issue mentioned above? Or we could let the user manage that.

mlt/utils/process_helpers.py
```diff
@@ -40,5 +42,17 @@ def run_popen(command, shell=False, stdout=PIPE, stderr=PIPE, cwd=None):
...
+            print("The following command is invalid:\n{}".format(command))
+            sys.exit(1)
+        elif isinstance(command, str):
+            command = command.split()
+        if False in map(lambda x: isinstance(x, str), command):
```
s1113950: I would comment this as saying `handles when something in a command array isn't a string` or something because it's not immediately clear what this check handles.

mlt/utils/process_helpers.py
```diff
@@ -40,5 +42,17 @@ def run_popen(command, shell=False, stdout=PIPE, stderr=PIPE, cwd=None):
...
+            print("The following command is malformed:\n{}".format(command))
+            sys.exit(1)
+        if binary_path(command[0]) is None:
```
s1113950: I dunno how valuable this is; my command could be `"cd .. && salad create` and I'll get an error that `salad` isn't a command since this just checks the first word in the command string. I'd just let `Popen` handle when the command isn't valid.

tests/unit/utils/test_process_helpers.py
```diff
@@ -69,8 +69,40 @@ def test_run_popen(popen):
...
+    popen.return_value = 0
+    result = run_popen(['ls', '/tmp'], shell=True)
+    assert result == 0
+
+
+@patch('mlt.utils.process_helpers.Popen')
+def test_run_popen_invalid_cmd(popen):
+    with pytest.raises(SystemExit) as pytest_raised_err:
+            run_popen(0)
+    assert pytest_raised_err.type == SystemExit
```
s1113950: No need for this because `with pytest.raises(SystemExit)` is already doing this check

tests/unit/utils/test_process_helpers.py
```diff
@@ -69,8 +69,40 @@ def test_run_popen(popen):
...
+    assert pytest_raised_err.value.code == 1
+
+
+@patch('mlt.utils.process_helpers.Popen')
+def test_run_popen_list_bad(popen):
+    with pytest.raises(SystemExit) as pytest_raised_err:
+            run_popen(['Python4', '--version'])
+    assert pytest_raised_err.type == SystemExit
```
s1113950: No need for this because `with pytest.raises(SystemExit)` is already doing this check

tests/unit/utils/test_process_helpers.py
```diff
@@ -69,8 +69,40 @@ def test_run_popen(popen):
...
+    assert pytest_raised_err.value.code == 1
+
+
+@patch('mlt.utils.process_helpers.Popen')
+def test_run_popen_malformed_cmd(popen):
+    with pytest.raises(SystemExit) as pytest_raised_err:
+            run_popen(['Python', 3, '--version'])
+    assert pytest_raised_err.type == SystemExit
```
s1113950: No need for this because `with pytest.raises(SystemExit)` is already doing this check

mlt/utils/process_helpers.py
ashahba: If I can't find a clean way to `validate invalid` commands, I'll take your suggestion. It makes sense 🙂

Dockerfile
ashahba: Removed.

mlt/utils/process_helpers.py
ashahba: I'd prefer to let user manage their `shell` arg, but the checks are still valid and required.
The only one that I have modify the the one in line 53 I guess as we talked offline.

mlt/utils/process_helpers.py
ashahba: This check is removed now.

PR #354, opened 2018-07-18
Horovod/Unet example using Horovod template

karthikvadla:

## Reviewer Checklist

 - [ ] Do you understand it?
 - [ ] Is it correct?
 - [ ] Is it safe?
 - [ ] Is it robust?
 - [ ] Is it simple?
 - [ ] Is it tested?
 - [ ] Is it documented?


examples/horovod_unet/.gitignore
```diff
@@ -0,0 +1,7 @@
+.build.json
+.push.json
+mlt.json
+*.swp
+.push.log
+.build.log
+.DS_Store
```
s1113950: is newline needed?

examples/horovod_unet/README.md
```diff
@@ -0,0 +1,27 @@
+# Distributed U-Net model training using horovod and openmpi on kubernetes
+
+This example explains the process of adapting an MLT horovod template to train
+a different model.  In this case, we are using the
+distributed U-Net](https://github.com/NervanaSystems/topologies/tree/master/distributed_unet/Horovod),
+which locates tumors in brain scan images.
+
+## Prerequisites
+
+This example requires that you have MLT and it's depedencies installed.
```
s1113950: No `'` needed

examples/horovod_unet/README.md
```diff
@@ -0,0 +1,27 @@
...
+* Please refer to this [section](https://github.com/IntelAI/mlt/tree/master/examples/distributed_unet#transferring-data-to-the-cluster)
+
+## Updating the app to run the U-Net model
+
+In order to create the Horovod/U-Net app for MLT, we started by initializing
+the app using `horovod` template.  We picked this template since
+it was most similar to what we needed to do for running the U-Net model.
+These are the steps that we took to update the app
+
+1.
```
s1113950: Need to add the steps here?

examples/horovod_unet/mlt_app/Dockerfile.cpu
```diff
@@ -0,0 +1,93 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+
+# Copyright (c) 2018 Intel Corporation
```
s1113950: Looks like you double-copied the copyright here

examples/horovod_unet/mlt_app/converter.py
```diff
@@ -0,0 +1,199 @@
...
+
+import os
+
+import nibabel as nib
+import numpy as np
+import numpy.ma as ma
+import settings
+from tqdm import tqdm
+
+root_dir = 'path_to_your_Brats_dir'  # Replace with your BraTS data directory
```
s1113950: Should this be replaced by an env var? How does this pass tests if it's hardcoded? 🤔

examples/horovod_unet/mlt_app/exec_multiworker.sh
```diff
@@ -0,0 +1,10 @@
+#! /bin/bash
+
+# Execute training on each node, pulling local thread counts for the train script
+cd /src/app
+
+export physical_cores=`lscpu | grep "Core(s) per socket" | cut -d':' -f2 | sed "s/ //g"` # Total number of physical cores per socket
+export num_threads=$(( ${2} * $physical_cores )) # Total number of physical cores on this machine
```
s1113950: Is this implying hyperthreading is enabled?

examples/horovod_unet/mlt_app/exec_multiworker.sh
```diff
@@ -0,0 +1,10 @@
+#! /bin/bash
+
+# Execute training on each node, pulling local thread counts for the train script
+cd /src/app
+
+export physical_cores=`lscpu | grep "Core(s) per socket" | cut -d':' -f2 | sed "s/ //g"` # Total number of physical cores per socket
```
s1113950: We should inform the user this doesn't work on OSX:
```
robertso@robertso-mac02:~/Nervana/fox-sentiment$ lscpu
-bash: lscpu: command not found
```
or you can do it the way we did in `Makefile`?
```
ifeq ($(OS), Darwin)
	NUMBER_OF_THREADS := $(shell sysctl -n hw.ncpu)
else
	NUMBER_OF_THREADS := $(shell nproc --all)
endif
```

examples/horovod_unet/mlt_app/converter.py
karthikvadla: I guess we don't have test cases around `examples`. Anyhow we are not using `converter.py` anywhere during training. This is to convert RAW images to `.npy` files. We are already using pre-processed data. 
This is just for a user help. I documented this in readme.

examples/horovod_unet/mlt_app/exec_multiworker.sh
karthikvadla: Updated/

examples/horovod_unet/mlt_app/exec_multiworker.sh
karthikvadla: @tonyreina should know better.

tonyreina: No hyperthreading would be logical cores not physical cores. I don't think we've had enough experience yet with this to experiment with the hyperthreading, but it's a good idea. The underlying TensorFlow script does take advantage of it if possible, but what we've been trying so far is just to pin the mpi threads to the physical cores.

ashahba: As long as total number of parallel threads for `mpi` are less or equal total number of cores across all cpus we should be good. So I guess in this case we want:
```
export num_threads=nproc
```

tonyreina: Yes. I think Ebi's way is a little cleaner.

I'm confused about the MaxOSX though. Isn't the container still an Ubuntu distro? I thought the container would be independent of the host OS so that we just needed to target the container OS not the host. So the `nproc` or `lscpu` command would still run within the container via Ubuntu (which is using the docker base, which is using the hypervisor).

karthikvadla: yeah containers are `ubuntu`. I think we should we be good .  I updated `lscpu` from `deploy.sh` with comments.

karthikvadla: So instead of these two lines
```export physical_cores=`lscpu | grep "Core(s) per socket" | cut -d':' -f2 | sed "s/ //g"` # Total number of physical cores per socket
export num_threads=$(( ${2} * $physical_cores )) # Total number of physical cores on this machine
```
we just need 
`export num_threads=$(nproc --all)` ?

tonyreina: That should work too. We just need a way for the bash script to probe for the number of cores on the machine. The easier and most generic way (i.e. works on the most OS) the better.

s1113950: ohh yeah if this script is ran in a container it should be fine to be this way

examples/horovod_unet/mlt_app/converter.py
s1113950: Why don't we have test cases around examples? 🤔 If they use `mlt` then couldn't we test them?

karthikvadla: nope.

PR #367, opened 2018-07-26
Peg ksync to version 0.3.1

ashahba:
## Reviewer Checklist

 - [x] Do you understand it?
 - [x] Is it correct?
 - [x] Is it safe?
 - [x] Is it robust?
 - [x] Is it simple?
 - [x] Is it tested?
 - [x] Is it documented?


scripts/ksync_install.sh
```diff
@@ -33,8 +33,7 @@ fi
 get_latest_release() {
```
s1113950: No need for this function now; it's not used anywhere but the old wget call

ashahba:
@s1113950 this should be good now and `ksync` install works again.

PR #374, opened 2018-07-31
Pep pip to version 9.0.3 until newer versions are functional.

### LINKED ISSUE #373: Peg pip to version 9.0.3 until newer versions are fixed.
Earlier today I noticed all of our builds started failing with due to:
```
py2-e2e create: /usr/share/mlt/.venv
py2-e2e installdeps: -rrequirements.txt, -rrequirements-dev.txt, -e.
ERROR: invocation failed (exit code 1), logfile: /usr/share/mlt/.venv/log/py2-e2e-1.log
ERROR: actionid: py2-e2e
msg: getenv
cmdargs: ['/usr/share/mlt/.venv/bin/pip', 'install', '-rrequirements.txt', '-rrequirements-dev.txt', '-e.']
```
Turns out this is a bug in `pip>=10.0.0` as described here:
https://github.com/scrapy/scrapy/issues/2115
https://github.com/retspen/webvirtcloud/issues/96

ashahba:
fixes #373 

## Reviewer Checklist

 - [ ] Do you understand it?
 - [ ] Is it correct?
 - [ ] Is it safe?
 - [ ] Is it robust?
 - [ ] Is it simple?
 - [ ] Is it tested?
 - [ ] Is it documented?


ashahba:
We also need `.circleci/config.yml` file to be updated since they just upgraded their images and that broke our builds:
https://hub.docker.com/r/circleci/python/tags/

ashahba:
@s1113950 I'm not sure what you mean 🙂

s1113950:
Your config file comment; at the top of the file we specify which circleci image we'd like to use. We could just use a new one (I think we were already using an old one).

ashahba:
Oh, I see. I guess I wasn't clean enough.
We are already using the new image: `circleci/python:2.7-stretch` which as of now is only `17 hours` old.

What I meant was they updated that image and all the other ones and that's why our builds started failing yesterday afternoon.

s1113950:
Ah ok. In that case can we use a python 3 image or an image that combines them? We had a python 3 image before so I'm not sure why we would move back to python 2?

ashahba:
Sure let me try that.

s1113950:
See my comments in the original issue. I don't think `pip` version is the issue here.

ashahba:
I'll give that a try shortly.

PR #377, opened 2018-08-03
Reduce the size of GKE cluster for CircleCi test runs and adjust pytest parallelism

ashahba:
Let's use `2` parallel threads for `e2e` tests and reduce cluster size until we actually have larger `CircleCi` VM, but I'm not sure if we even need that increase.

## Reviewer Checklist

 - [x] Do you understand it?
 - [x] Is it correct?
 - [x] Is it safe?
 - [x] Is it robust?
 - [x] Is it simple?
 - [x] Is it tested?
 - [x] Is it documented?


ashahba:
But first, let's fix `CircleCi` 🤞

PR #385, opened 2018-08-07
Ensure all runs have the same order so that the results are deterministic

ashahba:
Ensure all runs have the same order so that the results are deterministic

## Reviewer Checklist

 - [ ] Do you understand it?
 - [ ] Is it correct?
 - [ ] Is it safe?
 - [ ] Is it robust?
 - [ ] Is it simple?
 - [ ] Is it tested?
 - [ ] Is it documented?

### DIFF
--- a/tests/e2e/test_deploy_flow.py
+++ b/tests/e2e/test_deploy_flow.py
@@ -55,9 +55,9 @@ class TestDeployFlow(CommandTester):
     # TODO: find a way to test `tensorboard-bm`.
     # `tensorboard-bm` template requires the user domain to be configured
     # as a custom parameter https://github.com/IntelAI/mlt/issues/371
-    @pytest.mark.parametrize('template', filter(lambda x: os.path.isdir(
+    @pytest.mark.parametrize('template', sorted(filter(lambda x: os.path.isdir(
         os.path.join('mlt-templates', x)) and
-        ('tensorboard-bm' not in x), os.listdir('mlt-templates')))
+        ('tensorboard-bm' not in x), os.listdir('mlt-templates'))))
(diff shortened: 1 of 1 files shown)


tests/e2e/test_deploy_flow.py
```diff
@@ -55,9 +55,9 @@ def teardown(self):
     # TODO: find a way to test `tensorboard-bm`.
     # `tensorboard-bm` template requires the user domain to be configured
     # as a custom parameter https://github.com/IntelAI/mlt/issues/371
-    @pytest.mark.parametrize('template', filter(lambda x: os.path.isdir(
+    @pytest.mark.parametrize('template', sorted(filter(lambda x: os.path.isdir(
```
s1113950: This won't work with threaded tests. We determined it wasn't threading that was the problem I thought? Or are we disabling threading for now? If we're disabling them then can you add that to this PR?

ashahba: Yes, I think we are disabling them for now. Aren't we?

s1113950: @dmsuehir I've been doing other stuff and haven't fully been keeping up with the debugging on the current mlt issues; for sure threading is causing issues?

dmsuehir: No, I don't think we know for sure that threading is causing issues.  That horovod test was hanging in Ebi's PR and it was not threaded.

s1113950: @ashahba then we should close this PR, or you can keep it for now and disable threading here just as a debugging tool.

PR #386, opened 2018-08-08
Divide e2e tests into 2 groups

ashahba:
We can still keep `tf-dist-mnist` within first group since the running time is about as long as the whole set.

## Reviewer Checklist

 - [ ] Do you understand it?
 - [ ] Is it correct?
 - [ ] Is it safe?
 - [ ] Is it robust?
 - [ ] Is it simple?
 - [ ] Is it tested?
 - [ ] Is it documented?


dmsuehir:
Are you trying to split out the tests that might be more resource intensive to not be threaded?  We saw horovod hang a couple of times, even when it wasn't threaded, so I'm not sure that this will help?

PR #389, opened 2018-08-09
e2e test updates to check log content for `deploy -l`

### LINKED ISSUE #375: Sometimes after running log tailing e2e test, processes stay around
image

Note the 2 tailing pipe processes that are left over after `kubetail` runs.
Procs look like the line from around here: https://github.com/johanhaleby/kubetail/blob/master/kubetail#L280

We should ensure that these processes don't hang around as they use almost 1 core by themselves

dmsuehir:
fixes #375 

This code was originally added to the `pdb_debug` PR, but it didn't belong there and these changes seemed to be causing a hang, so there's another issue that needs to be debugged before this can be merged.

## Reviewer Checklist

 - [ ] Do you understand it?
 - [ ] Is it correct?
 - [ ] Is it safe?
 - [ ] Is it robust?
 - [ ] Is it simple?
 - [ ] Is it tested?
 - [ ] Is it documented?


s1113950:
I'll take this over since I wrote this stuff originally in the other PR

s1113950:
to repro a hanging `tail` call from `kubetail` after test runs:
```
EXTRA_ARGS=`$MLT_REGISTRY_AUTH_COMMAND` TESTOPTS="" PY=2 TESTFILES=tests/e2e/test_deploy_flow.py::TestDeployFlow::test_deploy_check_logs make test-e2e-no-docker
```

s1113950:
https://unix.stackexchange.com/questions/416150/make-tail-f-exit-on-a-broken-pipe quite possibly why this stays around at like 100% cpu. Normally a `tail` call doesn't require any resources, but when subproc is terminated we get this:
image 20

s1113950:
passes now, rerunning for verification

s1113950:
This passes but we can't run it 10X times in a row without it failing at least 1. Looks like the errors aren't its fault, but I'm not entirely sure of that. @dmsuehir what do you wanna do? Should we merge it or keep fixing stability issues here?

s1113950:
OOh actually prior fails relate to this I think: https://github.com/IntelAI/mlt/pull/396/files could be rate limiting issue. @dmsuehir I think this is G2G now then

dmsuehir:
These changes LGTM and as long as the it's not introducing more intermittent failures, then I think we are good to merge.  I can't approve this PR, because I'm the one that opened it.

tests/test_utils/e2e_commands.py
```diff
@@ -222,11 +222,43 @@ def deploy(self, no_push=False, interactive=False, retries=10,
...
+                       other proc
+                    """
+                    assert "Will tail" in output
+                    assert self.app_name in output
+                    break
+                except (AssertionError, IOError):
+                    # ignore if file doesn't exist yet (model not deployed yet)
+                    # also ignore if file created but not written to yet
+                    pass
+                if check != 599:
```
ashahba: I think we can remove this line and next one @s1113950 .
If not there we sleep one extra second on top of `600` seconds which is very unlikely since that is our upper bound and usually we have the logs before then.

s1113950: Why wouldn't we want to prevent sleeping 1 extra second? 😂 what's the downside?

ashahba: Oh sorry, just the `if` part I meant.

PR #394, opened 2018-08-14
Tracking multiple jobs: for mlt undeploy.

### LINKED ISSUE #318: Make mlt deploy keep track of multiple jobs
It's really common to want to run multiple training jobs at once in parallel. These may be related and fall under a single experiment, or they may span multiple experiments. `mlt` should be able to track any number of deployments.

I would propose creating subdirectories under the `k8s` directory that hold each deployment. `mlt` can still default to the most recent.

WafaaT:
fixes #318

## Reviewer Checklist

 - [ ] Do you understand it?
 - [ ] Is it correct?
 - [ ] Is it safe?
 - [ ] Is it robust?
 - [ ] Is it simple?
 - [ ] Is it tested?
 - [ ] Is it documented?


mlt/commands/undeploy.py
```diff
@@ -43,35 +42,51 @@ def action(self):
...
+                self._undeploy_all(namespace, jobs_list)
+            elif self.args.get('--job-name'):
+                job_name = self.args['--job-name']
+                if job_name in jobs_list:
+                    self._undeploy_job(namespace, job_name)
+                else:
+                    print("The job-name not found.")
+                    sys.exit(1)
+            else:
+                print("Multiple jobs are found under this application,"
```
dmsuehir: I think it would be better if you put the check for `if len(job_list) == 1` here instead of earlier.  The way that the code written now, if there is only one job deployed (say it's called job "X"), but the user does the command `mlt undeploy --job-name Y`, then job "X" is going to be undeployed, because the code isn't going to check the `--job-name` arg at all.

dmsuehir: Also, for this `print` message, rather than telling them to do `mlt undeploy -h`, we should just say that they either need to do `mlt undeploy --all` or specify a single job to undeploy using `mlt undeploy --job-name <job-name>`.  That's more straight forward for the user and our `-h` output isn't very pretty (if you do `mlt undeploy -h` it doesn't limit the help output to just `undeploy` options - it shows everything 😞  )

mlt/utils/files.py
```diff
@@ -39,3 +40,22 @@ def is_custom(target):
...
     return custom
+
+
+def get_deployed_jobs():
+    """get the list of the deployed jobs."""
+    return [d for d in os.listdir('k8s')
+            if os.path.isdir(os.path.join('k8s', d))]
+
+
+def create_job_subdir(job_name):
```
dmsuehir: Is there a reason why you created a util function for this?  It looks like we are only using it once in `deploy.py` and it doesn't seem like something that would need to be shared.

mlt/utils/files.py
```diff
@@ -39,3 +40,22 @@ def is_custom(target):
...
+    """create a sub-directory in k8s with the given job name."""
+    job_sub_dir = 'k8s/{}'.format(job_name)
+    if not os.path.exists(job_sub_dir):
+        os.makedirs(job_sub_dir)
+    return job_sub_dir
+
+
+def remove_job_dir(job_dir):
```
dmsuehir: This function also seems unnecessary since it's just wrapping one line of code?

mlt/commands/undeploy.py
```diff
@@ -43,35 +42,51 @@ def action(self):
...
+        else:
+            # Call the specified sub-command
+            if self.args.get('--all'):
+                self._undeploy_all(namespace, jobs_list)
+            elif self.args.get('--job-name'):
+                job_name = self.args['--job-name']
+                if job_name in jobs_list:
+                    self._undeploy_job(namespace, job_name)
+                else:
+                    print("The job-name not found.")
```
dmsuehir: It might be helpful here to also print out the `jobs_list`?  So that they know what their options are?

tests/e2e/test_deploy_flow.py
```diff
@@ -163,3 +165,15 @@ def test_debug_wrapper(self):
...
         self.verify_pod_status(expected_status="Failed")
+
+    def test_undeploy_all(self):
+        """test `mlt undeploy --all` command."""
+        self.init()
+        self.build()
+        self.deploy()
+        self.verify_pod_status()
+        self.status()
+        self.deploy()
```
s1113950: This should be a `--no-push` here so that it speeds the test up.

tests/e2e/test_deploy_flow.py
```diff
@@ -163,3 +165,15 @@ def test_debug_wrapper(self):
...
+        self.verify_pod_status()
+        self.status()
+        self.undeploy(all_jobs=True)
```
s1113950: I think this is fine :)
There isn't really a way to pass `all_jobs` to the autouse fixture unless you wanna get super fancy with it, so I'd say just have this as it'd be cleaner than the other way (detecting the name of the test and if it says `undeploy_all` then pass `all_jobs=True` to the `self.undeploy()` call there)

s1113950: on second thought, detecting name would be the way to go because of the outstanding bug of undeploying twice and failing... I'll fix that up next and then we can revert the name detection thing, or you can ignore that the tests are failing and keep going on your PR. WDYT?

WafaaT: to me, looks like the work around is a bit complicated, I'm thinking about a fix for the teardown itself. cannot it be fixed it just by adding an exception to the 2nd try https://github.com/IntelAI/mlt/blob/ab00e4b1ef00ccbade39503ad4fc3b4f68a66c89/tests/e2e/test_deploy_flow.py#L44 , I will try this out!

s1113950: 🤔 I don't think so, because what if the `undeploy` really did fail? We also wanna test the `undeploy` logic :/ We can't just ignore exceptions when `undeploy`ing, unless you wanna ignore for now the specific error that happens because of the existing `undeploy` bug.

mlt/commands/undeploy.py
```diff
@@ -43,35 +43,56 @@ def action(self):
...
+                sys.exit(1)
 
+    def _undeploy_all(self, namespace, jobs_list):
+        """undeploy all jobs."""
+        for job in jobs_list:
+            self._undeploy_job(namespace, job)
+
+    def _undeploy_job(self, namespace, job_name):
+        """undeploy the given job name"""
+        job_dir = "k8s/{}".format(job_name)
```
dmsuehir: Can you add a check if the directory exists, that way we (hopefully) don't have a problem with doing the undeploy multiple times.

WafaaT: This should be covered above in this section ```jobs_list = files.get_deployed_jobs()
        if not jobs_list:
            print("This app has not been deployed yet.")
            sys.exit(1) ```
unless it got deleted for any reason while the code is executed so, let me give it a try!

dmsuehir: Oh, you're right.  It would either hit this or the check for a valid job name.

dmsuehir:
Don't forget to update the `docs/features.md`

mlt/commands/undeploy.py
WafaaT: I also check the job name here, 
```                if job_name in jobs_list:
                    self._undeploy_job(namespace, job_name)
                else:
                    print('Job-name %s not found in: %s' % (job_name,
                                                            jobs_list))
                    sys.exit(1)```

File: docs/features.md [hunk 8]
```diff
@@ -166,13 +166,18 @@ was deployed for the current project directory.
...
 
-This command undeploys the last job that was deployed from the current
+This command undeploys the jobs that were deployed from the current
 project directory.  This frees resources in Kubernetes by deleting the
 associated job, pods, deployments, etc.
 
+| Option | Description | Default |
+|--------|-------------|---------|
+| `--all` | Undeploys all the jobs that were deployed from the current project directory.|
+| `--job-name=<name>` | The name of the job to undeploy. |
```
s1113950: Is there a way to get the list of jobs to choose from?

tests/test_utils/e2e_commands.py
```diff
@@ -264,6 +264,26 @@ def undeploy(self, all_jobs=False, job_name=None):
         elif job_name:
             command.append("--job-name={}".format(job_name))
         self._launch_popen_call(command)
+
+    # TODO: merge with undeploy(...) to use undeploy(all_jobs)=True
```
s1113950: TODO here

tests/test_utils/e2e_commands.py
```diff
@@ -264,6 +264,26 @@ def undeploy(self, all_jobs=False, job_name=None):
...
+    # in teardown()
+    def undeploy__for_test_teardown(self):
```
s1113950: Extra `_` here, function isn't being called most likely?

tests/test_utils/e2e_commands.py
```diff
@@ -264,6 +264,26 @@ def undeploy(self, all_jobs=False, job_name=None):
...
+        p = run_popen(command, shell=False, stdout=PIPE,
+                      stderr=PIPE, cwd=cwd)
+        out, err = p.communicate()
+        error_msg = "Popen call failed:\nSTDOUT:{}\n\nSTDERR:{}".format(
+            str(out), colored(str(err), 'red'))
+        if p.returncode == 1:
+            # in case of no jobs undeploy, where the
+            # `mlt undeploy --all` returns this
+            # 'This app has not been deployed yet.' as output.
+            assert p.wait() == 1, error_msg
```
s1113950: `returncode` is the same as what `p.wait()` returns, so there's no need to check it twice.
This fix would be for the tests themselves but not for calling `mlt undeploy` repeatedly 🤔 @dmsuehir I thought we were gonna put the fix to that in `undeploy.py`?

docs/features.md
WafaaT: `mlt status` should do so, I'm working on it now.

dmsuehir: yeah, @WafaaT will update `mlt status` to show it, but she's developing this in phases, so this is just the first phase.  Next phase also has to update status, log, etc.

s1113950: 👍 🤔 starting with `undeploy` then seems a bit odd because if this lands by itself the end user can't really utilize it?

tests/test_utils/e2e_commands.py
WafaaT: I thought `p.wait()` returns both of `returncode` and `error_msg`, am I correct?  ..and I needed to check if the code is `1` code first. do you think `if p.wait() == 1: ` is correct?

dmsuehir: `undeploy.py` will no longer barf with a stack trace if we try to undeploy multiple times, but there is still a message that says something like "this app hasn't been deployed" (if there are no deployments) or "the job is not found" (if they specify a job name that doesn't exist in the deployments - which will also happen if they undeploy twice).  And Wafaa is doing a system.exit(1) in those cases, so will still hit this case.  

@WafaaT what if we check the output for those known error messages here, so that if something unexpected goes wrong, the test will fail.

@s1113950 what do you think?

tests/test_utils/e2e_commands.py
WafaaT: I fixed it in the new commit!

tests/test_utils/e2e_commands.py
s1113950: 👍 checking for specific errors works for me

tests/test_utils/e2e_commands.py
WafaaT: there's TODO already?!

tests/test_utils/e2e_commands.py
s1113950: ahh ok, I missed the change to `undeploy.py`. The `view new changes` button didn't show me that stuff so I must have looked at it by accident or something 🤔

s1113950: In that case I'm gonna do a full review just to see if I missed something!

tests/test_utils/e2e_commands.py
s1113950: :lol: you added this change? It's in your PR

tests/unit/commands/test_undeploy.py
```diff
@@ -68,67 +63,184 @@ def get_sync_spec_mock(patch):
...
+    remove_job_dir_mock.input_value = 'k8s/job1'
+    get_deployed_jobs_mock.return_value = {"job1"}
+    command = {'undeploy': True, '--job-name': 'job2'}
+    with catch_stdout() as output:
+        with pytest.raises(SystemExit):
+            UndeployCommand(command).action()
+        assert"Job-name job2 not found" in output.getvalue()
+
+
+def test_undeploy_synced(colored_mock,
```
s1113950: Same thing here possibly (if get_sync_spec_mock fits on 1 line)

docs/features.md
WafaaT: `mlt undeploy` still works as the user expects.. otherwise, the user can definitely use the new `mlt undeploy ....` with subcommands along with `kubctl get jobs --namespace... ` to  display the deployed jobs, just for the end of this sprint!

tests/test_utils/e2e_commands.py
WafaaT: @dmsuehir I can add a check for the specific output message here, in this case it was `this app hasn't been deployed`, is this what you mean? if I added this check then the `teardown` would be successfully executed in case of underlying jobs correctly and in case of no jobs and caught system exit 1 with this message "this app hasn't been deployed" ! 
for other test that uses `undeploy()` to test undpeloy by `job-name` I can add more tests later on once `mlt status` and validate them case by case.
but unit tests already covers most of the new changes.

mlt/commands/undeploy.py
```diff
@@ -43,35 +43,56 @@ def action(self):
...
+            elif self.args.get('--job-name'):
+                job_name = self.args['--job-name']
+                if job_name in jobs_list:
+                    self._undeploy_job(namespace, job_name)
+                else:
+                    print('Job-name %s not found in: %s' % (job_name,
+                                                            jobs_list))
+                    sys.exit(1)
+            elif len(jobs_list) == 1:
+                self._undeploy_job(namespace, jobs_list.pop())
```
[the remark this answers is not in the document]
s1113950:
```
In [1]: def foo():
   ...:     a = ['elem']
   ...:     a.pop()
   ...:

In [2]: def bar():
   ...:     a = ['elem']
   ...:     a[0]
   ...:

In [3]: import dis

In [4]: dis.dis(foo)
  2           0 LOAD_CONST               1 ('elem')
              2 BUILD_LIST               1
              4 STORE_FAST               0 (a)

  3           6 LOAD_FAST                0 (a)
              8 LOAD_ATTR                0 (pop)
             10 CALL_FUNCTION            0
             12 POP_TOP
             14 LOAD_CONST               0 (None)
             16 RETURN_VALUE

In [5]: dis.dis(bar)
  2           0 LOAD_CONST               1 ('elem')
              2 BUILD_LIST               1
              4 STORE_FAST               0 (a)

  3           6 LOAD_FAST                0 (a)
              8 LOAD_CONST               2 (0)
             10 BINARY_SUBSCR
             12 POP_TOP
             14 LOAD_CONST               0 (None)
             16 RETURN_VALUE
```
```
In [7]: timeit.timeit(foo)
Out[7]: 0.26696512200032885

In [8]: timeit.timeit(bar)
Out[8]: 0.17998152799918898
```
calling `[0]` is quite a bit faster than calling `.pop()` :)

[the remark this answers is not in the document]
ashahba: It's them the magic of `Array` vs `List`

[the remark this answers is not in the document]
s1113950: ??? Python has no concept of `Arrays` @ashahba . Thing here is calling `pop()` calls a function that at worst case needs to move every element ahead in the list under the hood (python lists are dynamic C arrays) so it should be slower. In this case where there's only 1 element, python probably has built in interpretation of `[0]` access vs having to call the `pop()` function in the other case

tests/test_utils/e2e_commands.py
```diff
@@ -293,8 +293,33 @@ def logs(self):
...
-        self._launch_popen_call(['mlt', 'undeploy'])
+    def undeploy(self, all_jobs=False, job_name=None):
+        command = ['mlt', 'undeploy']
+        if all_jobs:
+            command.append("--all")
+        elif job_name:
+            command.append("--job-name={}".format(job_name))
+        self._launch_popen_call(command)
+
+    # TODO: merge with undeploy(...) to use undeploy(all_jobs)=True
```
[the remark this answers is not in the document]
WafaaT: ??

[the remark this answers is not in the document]
s1113950: You added this TODO in your PR, shouldn't it get done before merge? There's effectively 2 `undeploy` functions in the case where this TODO doesn't get done. @dmsuehir am I missing something? :(

mlt/commands/undeploy.py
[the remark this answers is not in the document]
ashahba: Correct and that's the point because we are actually trying to drop first element of an underlying `array` here vs just trying to access it with `[0]` and we know which one is going take long specially if the underlying array is big 🙂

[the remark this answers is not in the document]
dmsuehir: I vote `jobs_list[0]`

mlt/commands/undeploy.py
```diff
@@ -43,35 +43,56 @@ def action(self):
...
+            else:
+                print("Multiple jobs are found under this application,"
+                      "please try `mlt undeploy --all` or specify a single"
+                      " job to undeploy using "
+                      "`mlt undeploy --job-name <job-name>`")
+                sys.exit(1)
 
+    def _undeploy_all(self, namespace, jobs_list):
+        """undeploy all jobs."""
+        for job in jobs_list:
```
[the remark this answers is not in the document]
dmsuehir: good question!  I just looked this up and it looks like it might be possible if you do the delete with `--recursive`.  @WafaaT want to try this?

```
  -R, --recursive=false: Process the directory used in -f, --filename recursively. Useful when you want to manage
related manifests organized within the same directory.
```

tests/unit/commands/test_undeploy.py
```diff
@@ -68,67 +63,184 @@ def get_sync_spec_mock(patch):
...
+    os_path_exists_mock.return_value = True
+    remove_job_dir_mock.input_value = 'k8s/job1'
+    with catch_stdout() as output:
+        UndeployCommand({'undeploy': True, '--all': True}).action()
+        assert output.getvalue().strip() == 'Successful Custom Undeploy' \
+                                            '\nSuccessful Custom Undeploy'
 
 
-def test_undeploy(load_config_mock, proc_helpers):
+def test_undeploy(load_config_mock,
```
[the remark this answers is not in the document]
dmsuehir: @WafaaT can you put it back to follow the pep8 standards?  If we decided to deviate from this for some reason, we'd want to do it for everything, not just one file.

mlt/commands/undeploy.py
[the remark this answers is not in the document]
WafaaT: I'm sure that I have a single element in the list, I'll change it to `jobs_list[0]` though  :)

tests/e2e/test_deploy_flow.py
```diff
@@ -104,6 +104,7 @@ def test_no_push_deploy(self):
         self.deploy()
         self.verify_pod_status()
         self.status()
+        self.undeploy()
```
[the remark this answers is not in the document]
WafaaT: yes, it would work too

tests/unit/commands/test_undeploy.py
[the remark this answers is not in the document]
WafaaT: I didn't knew we use pep8 standards, I will modify them!

tests/test_utils/e2e_commands.py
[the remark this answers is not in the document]
dmsuehir: I discussed with @WafaaT.  This is going to take time to refactor, so she will hold off on this change and file an MLT issue to come back and clean up the undeploy/teardown functions later.

mlt/utils/files.py
```diff
@@ -39,3 +39,9 @@ def is_custom(target):
                     custom = True
                     break
     return custom
+
+
+def get_deployed_jobs():
+    """get the list of the deployed jobs."""
+    return [d for d in os.listdir('k8s')
```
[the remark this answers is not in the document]
WafaaT: chatted about this, agreed to skip it!

mlt/commands/deploy.py
```diff
@@ -227,17 +227,32 @@ def _custom_deploy(self, app_name, app_run_id, remote_container_name):
...
+
+    def _track_deployed_job(self, app_name, app_run_id):
+        """create a subdirectory in k8s with the deployed job name."""
+        job_name = "-".join([app_name, app_run_id])
+        return self.create_job_subdir(job_name)
+
+    def create_job_subdir(self, job_name):
+        """create a sub-directory in k8s with the given job name."""
+        job_sub_dir = 'k8s/{}'.format(job_name)
+        if not os.path.exists(job_sub_dir):
```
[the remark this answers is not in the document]
WafaaT: chatted about this, agreed to skip it!

mlt/commands/undeploy.py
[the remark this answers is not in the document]
WafaaT: it would remove `k8s` itself if I use `shut.rmtree('k8s')` it deletes all subdirs and `k8s`! , @dmsuehir I'm not sure how to use `--recursive=false` here. do you mean to use something like that `os.system("rm -rf %s" %filename))` ??
I need to keep `k8s` after deleting it contents, right?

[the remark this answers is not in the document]
dmsuehir: I meant doing something like `kubectl delete -f k8s --recursive`.  I thought that is what @s1113950 was referring to (doing the k8s undeploy on everything in the directory at once when we do undeploy all).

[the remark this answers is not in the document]
dmsuehir: I haven't tried doing the kubectl recursive delete, but the documentation I pasted in the earlier comment is from the kubectl help, so it seems like it might be possible.

s1113950:
There are some TODOs and things still but will fix those up in another PR coming soon

PR #395, opened 2018-08-15
Using intelaipg/intel-optimized-tensorflow:1.9.0-mkl-py3 images across the board - part 1

ashahba:
## Reviewer Checklist

 - [ ] Do you understand it?
 - [ ] Is it correct?
 - [ ] Is it safe?
 - [ ] Is it robust?
 - [ ] Is it simple?
 - [ ] Is it tested?
 - [ ] Is it documented?


mlt-templates/tf-dist-mnist/requirements.txt
```diff
@@ -1,3 +1,4 @@
 # Please don't add tensorflow here since it's being installed by the Docker
+# Installing Tensorflow due to: https://github.com/tensorflow/tensorflow/issues/20833
+tensorflow==1.8.0
```
s1113950: You're using the intel tensorflow image for this though. On the other templates you say to not install tensorflow here. Why is this any different?

ashahba: Yes, for now. But this is why:
https://github.com/kubeflow/tf-operator/issues/774

s1113950: So then does it make sense to not start from the Intel image and just use python3.6 for now?

ashahba: I have no strong opinion there, but I think using `Intel` images that has all the pieces in place including `MKL` and `Tensorflow` which already works with other templates is nice and reduces custom configurations required if we were to enable `MKL` ourselves.

karthikvadla:
Really good to have :)

mlt-templates/tf-dist-mnist/requirements.txt
s1113950: 👍 I'm cool with that

PR #398, opened 2018-08-16
Horovod Example: Add tensorboard support

karthikvadla:

## Reviewer Checklist

 - [ ] Do you understand it?
 - [ ] Is it correct?
 - [ ] Is it safe?
 - [ ] Is it robust?
 - [ ] Is it simple?
 - [ ] Is it tested?
 - [ ] Is it documented?


examples/horovod_unet/mlt_app/Dockerfile.cpu
```diff
@@ -72,6 +72,11 @@ RUN cat /etc/ssh/ssh_config | grep -v StrictHostKeyChecking > /etc/ssh/ssh_confi
     echo "    StrictHostKeyChecking no" >> /etc/ssh/ssh_config.new && \
     mv /etc/ssh/ssh_config.new /etc/ssh/ssh_config
 
+# TODO:
+# Check on how to inject envs after container is built and during deploy
+# We have to BASE_DIR and OUTPUT_DIR which are used in settings.py
```
s1113950: Should this be `we have to set`? `BASE_DIR` and `OUTPUT_DIR` are quite confusing verbs 😂

ashahba: `We have to set environment variables BASE_DIR and OUTPUT_DIR which are used in settings.py, but right now we have to set those manually.`

s1113950: @karthikvadla ^^^ is better than keeping the lines separate

ashahba: `We have to set environment variables DATA_PATH and OUTPUT_PATH which are used in settings.py, but right now we have to set those manually.`

examples/horovod_unet/mlt_app/main.py
```diff
@@ -0,0 +1,229 @@
...
+        hvd.init()
+
+    # Define model
+    model = define_model(imgs_train.shape, msks_train.shape, FLAGS)
+
+    if not FLAGS.no_horovod:
+        # Horovod: adjust learning rate based on number workers
+        opt = tf.train.AdamOptimizer(learning_rate=FLAGS.learningrate,
+                                      epsilon=tf.keras.backend.epsilon())
+        #opt = tf.train.RMSPropOptimizer(0.0001 * hvd.size())
```
ashahba: Do we still need these lines?

examples/horovod_unet/mlt_app/main.py
```diff
@@ -0,0 +1,229 @@
...
+        # tf.logging.info("HOROVOD: New learning rate is {}".\
+        #         format(FLAGS.learningrate * hvd.size()))
+    else:
+        opt = tf.train.AdamOptimizer(learning_rate=FLAGS.learningrate,
+                                      epsilon=tf.keras.backend.epsilon())
+        #opt = tf.train.RMSPropOptimizer(0.0001)
```
ashahba: same here

examples/horovod_unet/mlt_app/main.py
```diff
@@ -0,0 +1,229 @@
...
+                                           config=config) as mon_sess:
+
+        while not mon_sess.should_stop():
+
+            # Run a training step synchronously.
+            image_, mask_ = get_batch(imgs_train,
+                                      msks_train,
+                                      FLAGS.batch_size)
+
+            # Do batch in order
```
ashahba: same here

examples/horovod_unet/mlt_app/model.py
```diff
@@ -0,0 +1,216 @@
...
+    summary_op = tf.summary.merge_all()
+
+    return model
+
+
+def validate_model(mon_sess, model, imgs_test, msks_test):
+    """
+    Code for model validation on test set
+    """
+    # test_dice, test_sens, test_spec = mon_sess.run(
```
ashahba: same here

examples/horovod_unet/mlt_app/settings.py
```diff
@@ -20,30 +20,29 @@
...
 # Mode 3: Use T2 to identify the active core (necrosis, enhancing, non-enh)
 MODE = 1  # 1, 2, or 3
 
-import psutil
-
 BLOCKTIME = 0
 NUM_INTER_THREADS = 2
-NUM_INTRA_THREADS = psutil.cpu_count(logical=False)  # Total number of physical cores across all sockets
+# Total number of physical cores across all sockets
+NUM_INTRA_THREADS = 72
```
ashahba: Did we want this to be hard coded? That's a lot of threads 🤔

examples/horovod_unet/mlt_app/deploy.sh
```diff
@@ -74,21 +74,20 @@ NODE_SELECTOR="node-type=highmem"
...
-LOG_DIR=${LOG_DIR}
 WORKERS=$(( ${NUM_NODES} * ${NUM_WORKERS_PER_NODE} ))
 SOCKETS_PER_NODE=${SOCKETS_PER_NODE}
 NUM_INTER_THREADS=${NUM_INTER_THREADS}
 
 PPR=$(( $NUM_WORKERS_PER_NODE / $SOCKETS_PER_NODE ))
 PE=$(( $PHYSICAL_CORES / $PPR ))
 GPU=${GPUS}
-EXEC="mpirun -np ${WORKERS} --hostfile /kubeflow/openmpi/assets/hostfile --map-by socket -cpus-per-proc $PHYSICAL_CORES --report-bindings --oversubscribe bash /src/app/exec_multiworker.sh ${LOG_DIR} ${PPR} ${NUM_INTER_THREADS}"
+EXEC="mpirun -np ${WORKERS} --hostfile /kubeflow/openmpi/assets/hostfile --map-by socket -cpus-per-proc $PHYSICAL_CORES --report-bindings --oversubscribe bash /src/app/exec_multiworker.sh ${PPR} ${NUM_INTER_THREADS}"
```
ashahba: Don't we need `LOG_DIR` anymore?

examples/horovod_unet/mlt_app/Dockerfile.cpu
ashahba: Luckily they are `DATA_PATH` and `OUTPUT_PATH` now.

examples/horovod_unet/mlt_app/deploy.sh
karthikvadla: nope, i renamed it to `OUTPUT_PATH`

examples/horovod_unet/mlt_app/settings.py
karthikvadla: I got those updated scripts from tony, i ran those 4-node. Training completed successfully

examples/horovod_unet/mlt_app/Dockerfile.cpu
karthikvadla: yeah fixing that

examples/horovod_unet/mlt_app/settings.py
ashahba: Sure, but I mean if want people to use the example we have to give them heads up that this default value needs to be carefully reviewed since not everyone have access to those many cores.

examples/horovod_unet/mlt_app/main.py
karthikvadla: I will leave those as it is, i'm not sure those are left intentionally

tonyreina: I'd keep them. I was trying to follow one of the Horovod examples, but found that it was better not to adjust the learning rate with the number of workers.

examples/horovod_unet/mlt_app/model.py
tonyreina: Yeah. The idea here is to replicate the tf-dist-mnist example. I've refactored the code to just have a `main.py`, `data.py`, and `model.py`.  So `main.py` probably won't get changed. However, if the user had a different dataset, then `data.py` would be modified to the new data loader. And, if they had a new model, then `model.py` would get modified. This way it is clear to the DS what code they need to change and where to look for it.

examples/horovod_unet/mlt_app/settings.py
tonyreina: No, the threads should be based on the number of physical cores. So we should change those back to the psutil.cpu_count(logical=False).  This way the system gets probed for how many cores it has.

ashahba: Yes to echo what @tonyreina mentioned in general anytime you ask for more resources available to you, your job idle fail or worse hang indefinitely 🙂

examples/horovod_unet/mlt_app/mlt.json
```diff
@@ -11,6 +11,6 @@
     "data_path": "/var/datasets/unet/vck-resource-8fd827a8-809a-11e8-b982-0a580a480bd4",
     "output_path": "/var/datasets/unet/vck-resource-8fd827a8-809a-11e8-b982-0a580a480bd4"
   },
-  "registry": "<docker_io>or<gcr.io>",
+  "registry": "gcr.io/constant-cubist-173123",
```
ashahba: I would remove the project name here and make it something like:
```
"registry": "YOUR_REGISTRY",
```

examples/horovod_unet/mlt_app/settings.py
tonyreina: Yes. Not sure if we want to tweak that. If you are the sole occupant on the node, then it is better to ask for all physical cores (for speed). However, maybe there is a K8s-friendly way to do this?? I don't know. The `psutil` code just probes it at runtime.

ashahba: I guess the point I'm raising is: What if the aggregated number of physical cores across the nodes where the Pods are being deployed is less than `72`? Does this fails? or hang indefinitely? As long as it fails with an error showing why it failed we should be ok.

examples/horovod_unet/mlt_app/main.py
```diff
@@ -0,0 +1,233 @@
...
+
+import settings
+
+
+FLAGS = tf.app.flags.FLAGS
+
+tf.app.flags.DEFINE_integer("num_inter_threads", 2,
+                            "# inter op threads")
+
+tf.app.flags.DEFINE_integer("num_threads", psutil.cpu_count(logical=False),
```
karthikvadla: @ashahba here.

karthikvadla:
@ashahba Addressed your comment, removed hard coding `NUM_INTRA_THREADS`, because we are calculating programmatically now.

PR #400, opened 2018-08-16
Added linux prereq install script

s1113950:

## Reviewer Checklist

 - [ ] Do you understand it?
 - [ ] Is it correct?
 - [ ] Is it safe?
 - [ ] Is it robust?
 - [ ] Is it simple?
 - [ ] Is it tested?
 - [ ] Is it documented?


s1113950:
🤔 should I just add a mac install script here too? Most of our devs are on macs. @dmsuehir

dmsuehir:
@s1113950 The mac one can be a separate PR if you want.  It's awesome that we can test the linux install script as part of our test cluster setup.  How will we test the mac one?

scripts/linux_prereq_install.sh
```diff
@@ -0,0 +1,22 @@
+#! /bin/bash
+
+# TODO: is this even necessary? Users should have `git` already
+# jq is needed for tensorboard
+# openssh is needed for horovod
+apt-get install git jq openssh-server openssh-client
```
ashahba: Because of this line, should we call the file `ubuntu_prereq_install.sh` ?

s1113950:
Unfortunately it won't be tested regularly :( I'd add a `DISCLAIMER` at the top of the file. We would need a `mac` setup on circleci which would be a paying feature I thought

scripts/linux_prereq_install.sh
s1113950: I could rename it to `debian_prereq_install`? `apt` is a debian-based installer :)

ashahba: Sounds good.

ashahba: We could also use this pattern:
https://github.com/IntelAI/mlt/blob/f4775e8a778dfadbd0cb51c6a2c1d79198292a8c/scripts/ksync_install.sh#L26
to make it both `MacOS` and `Linux` compatible, but I don't know if you want to do it for this ticket.

Makefile
```diff
@@ -144,6 +144,10 @@ docker:
 		--build-arg no_proxy=${no_proxy} \
 		-t mlt .
 
+# useful for linux users to get every outside dependency required for mlt to work
+linux_prereq_install:
+	@./scripts/linux_prereq_install.sh
```
dmsuehir: change to`debian_prereq_install` here too

PR #407, opened 2018-08-22
Adds developer workspace config

### LINKED ISSUE #74: Enable developer/workspace config of container registry and template repository
Currently, `mlt init` is getting a bit heavy in terms of arguments needed to be provided when custom container registry and template repositories are used.
It would be handy if you could predefine those in either environment variables or in a config file.

s1113950:
fixes #74 

## Reviewer Checklist

 - [ ] Do you understand it?
 - [ ] Is it correct?
 - [ ] Is it safe?
 - [ ] Is it robust?
 - [ ] Is it simple?
 - [ ] Is it tested?
 - [ ] Is it documented?

It does this via env vars that a user can set.
Env vars are in this format:
`MLT_{VAR}`
where `VAR` is any flag they wanna set. `_` are converted to `-` in things like `MLT_SKIP_CRD_CHECK`

Order of precedence:
`command-line vars` --> `template config` --> `Env vars` -->


docs/features.md
```diff
@@ -25,6 +25,18 @@
...
 
+## Global Config
+
+Environment variables will be used in place of `mlt` flags if named as below, with `MLT_` followed by the name of the flag. `-` are converted to `_`.
+
+Examples: 
+
+`MLT_NAMESPACE`: `--namespace`
+`MLT_TEMPLATE_REPO`: `--template-repo`
+`MLT_SKIP_CRD_CHECK`: `--skip-crd-check`
```
dmsuehir: I think registry will be a common thing that people want to set as a global config, so maybe we mention that as an example here?

mlt/main.py
```diff
@@ -108,7 +109,7 @@
 # every available command and its corresponding action will go here
 COMMAND_MAP = (
     ('build', BuildCommand),
-    ('config', ConfigCommand),
+    ('template_config', ConfigCommand),
```
dmsuehir: Maybe we should we rename `ConfigCommand` to `TemplateConfigCommand` for consistency?

dmsuehir:
Do a global search for `mlt config` because I'm noticing it in some other places like the `tensorboard-bm` and `tensorboard-gke` readme files.

Issue #436, opened 2018-09-14
mlt undeploy backwards compatibility is broken


s1113950:
```
robertso@robertso-mac02:/tmp/wat$ mlt undeploy --job-name wat-60583ba9-93bc-43c5-ac53-602a6a09593a
robertso@robertso-mac02:/tmp/wat$ mlt status
Job: wat-89676d81-78d8-43c2-bd12-efff4c10652f -- Creation Time: 2018-09-14 21:30:50+00:00
No resources found.
```
```
robertso@robertso-mac02:/tmp/wat$ mlt undeploy
Error from server (NotFound): error when stopping "k8s/wat-89676d81-78d8-43c2-bd12-efff4c10652f/job.yaml": jobs.batch "wat-89676d81-78d8-43c2-bd12-efff4c10652f" not found
Traceback (most recent call last):
  File "/Users/robertso/Nervana/mlt/.venv3/bin/mlt", line 11, in <module>
    load_entry_point('mlt==0.3.0+3.g5c5c414', 'console_scripts', 'mlt')()
  File "/Users/robertso/Nervana/mlt/mlt/main.py", line 213, in main
    run_command(args)
  File "/Users/robertso/Nervana/mlt/mlt/main.py", line 134, in run_command
    CommandClass(args).action()
  File "/Users/robertso/Nervana/mlt/mlt/commands/undeploy.py", line 60, in action
    self._undeploy_jobs(namespace, jobs, all_jobs=True)
  File "/Users/robertso/Nervana/mlt/mlt/commands/undeploy.py", line 86, in _undeploy_jobs
    raise_on_failure=True)
  File "/Users/robertso/Nervana/mlt/mlt/utils/process_helpers.py", line 31, in run
    raise e
  File "/Users/robertso/Nervana/mlt/mlt/utils/process_helpers.py", line 28, in run
    output = check_output(command, cwd=cwd).decode("utf-8")
  File "/usr/local/Cellar/python/3.6.4_3/Frameworks/Python.framework/Versions/3.6/lib/python3.6/subprocess.py", line 336, in check_output
    **kwargs).stdout
  File "/usr/local/Cellar/python/3.6.4_3/Frameworks/Python.framework/Versions/3.6/lib/python3.6/subprocess.py", line 418, in run
    output=stdout, stderr=stderr)
[...]

PR #441, opened 2018-10-01
fixed undeploy backwards compatibility and added helper func to catch errors

s1113950:
fixes https://github.com/IntelAI/mlt/issues/436
Also refactored how we catch errors a bit

We now have `error_handling.throw_error` function to use instead of manually printing and then exiting

## Reviewer Checklist

 - [ ] Do you understand it?
 - [ ] Is it correct?
 - [ ] Is it safe?
 - [ ] Is it robust?
 - [ ] Is it simple?
 - [ ] Is it tested?
 - [ ] Is it documented?


mlt/utils/error_handling.py
```diff
@@ -0,0 +1,26 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+import sys
+
+
+def throw_error(error):
```
dmsuehir: Should we do the red color thing as a parameter here too?  That would save us from having to import the colored library everywhere else.  And since we now have this error handler helper thing we should convert other code to use it too for consistency.

tests/unit/commands/test_init.py
```diff
@@ -193,8 +203,7 @@ def test_init(open_mock, process_helpers, copytree_mock, check_output_mock,
 def test_init_enable_sync(open_mock, process_helpers, copytree_mock,
                           check_output_mock, config_helpers_mock,
                           copyfile_mock, listdir_mock,
-                          binary_path_mock, os_path_exists_mock,
-                          update_yaml_for_sync_mock):
+                          binary_path_mock, os_path_exists_mock):
```
dmsuehir: did you remove the update_yaml_for_sync_mock?  because I still see it being used on line 221

s1113950: It's used for another test but shouldn't be used for this one

PR #464, opened 2018-10-26
adding W503 to ignore list for pep8 and also fix W504 and W605 errors

ashahba:
We are getting a bunch of style errors since W504 is enforced by default now and W503 was supposed to added to ignore list by default but that doesn't seem to work.

So this is what this fix is about.

Read more about it here: https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator


ashahba:
I can skip the examples like we used to, that should be fine @dmsuehir

ashahba:
@dmsuehir I added `examples` folder to the ignore list for style checking 🙂

dmsuehir:
@ashahba weird, I thought it was already there.  Thanks

ashahba:
Only [examples/horovod_unet](https://github.com/IntelAI/mlt/pull/464/files#diff-b91f3d5bd63fcd17221b267e851608e8L31) was in the list 🙂

